PR #11, opened 2020-04-30
Support AbortController / AbortSignal

addaleax:
Support AbortController, as well as a Node.js-`EventEmitter`-y variant
of it for people who don’t want to pull in external dependecies or
write their own implementation of it.

Fixes: https://github.com/jasnell/piscina/issues/7

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ This class extends [`EventEmitter`][] from Node.js.
   * `useAtomics`: (`boolean`) Use the [`Atomics`][] API for faster communication
     between threads. This is on by default.
 
-### Method: `runTask(task[, transferList][, filename])`
+### Method: `runTask(task[, transferList][, filename][, abortSignal])`
 
 Schedules a task to be run on a Worker thread.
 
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "build": "tsc",
     "lint": "standardx \"**/*.ts\" | snazzy",
     "test": "npm run lint && npm run build && npm run test-only",
-    "test-only": "cross-env TS_NODE_PROJECT=test/tsconfig.json tap --ts --no-browser --coverage-report=html --coverage-report=text --100 test/*.ts",
+    "test-only": "cross-env TS_NODE_PROJECT=test/tsconfig.json tap --ts --no-browser --coverage-report=html --coverage-report=text --100 --jobs=2 test/*.ts",
```
addaleax, line 12: This is limiting concurrency, because every single one of our tests spawns a pool of Workers and running them with full parallelism makes me hit memory limits locally :smile:

src/index.ts
```diff
@@ -352,7 +384,24 @@ class ThreadPool {
         } else {
           resolve(result);
         }
+      },
+      abortSignal);
+
+    if (abortSignal !== null) {
+      onabort(abortSignal, () => {
+        reject(new AbortError());
```
mcollina, line 392: please add a comment that we do not return because we need to do more work about the rejection.

package.json
mcollina: I would recommend adding in a `.taprc` file. We start to have a bit too many options :P

README.md
```diff
@@ -79,10 +79,17 @@ Schedules a task to be run on a Worker thread.
...
+  cancel a task, including already running ones.
+  (More generally, any `EventEmitter` or `EventTarget` that emits `'abort'`
+  events can be passed here.) Abortable tasks cannot share threads regardless
+  of the `concurrentTasksPerWorker` options.
 
 This returns a `Promise` for the return value of the (async) function call
 made to the function exported from `filename`. If the (async) function throws
 an error, the returned `Promise` will be rejected with that error.
+If the task is aborted, the returned `Promise` is rejected with an error
+as well.
```
mcollina, line 93: Can you please expand to what happens to the worker if the task is aborted while it was running?

package.json
addaleax: Done :+1:

README.md
addaleax: Done, but imo it’s an implementation detail that we should not focus on here.

src/index.ts
addaleax: I’ve added a comment but I’m not sure if that’s what you hand in mind (what does “do more work about the rejection” mean?)

PR #12, opened 2020-04-30
scrypt example

jasnell:
@addaleax @mcollina ... this is just a start of an example workload so we can start refining the ideal kinds of workloads for pooling.

```
$ cd examples/scrypt

$ npm run pooled 
// Runs the sample scrypt'ing 1000 randomly generated byte arrays produced
// using an async generator and fed into piscina tasks with the default min/max
// workers. This uses the async version of scrypt and randomFill, which means we're
// also bound to the libuv threadpool.

$ npm run pooled-sync
// Same as above except the sync version of scrypt is used within the worker tasks.

$ npm run unpooled
// Same as pooled except that piscina is not used at all

$ npm run unpooled-sync
// Same as pooled-sync except that piscina is not used at all
```

Each run will output the duration measured with `performance.mark`/`performance.measure`

This isn't necessarily an example we should keep, but it does provide a starting point for reasoning about the kinds of workloads that we should highlight. Please knock this over as much as possible

### DIFF
--- a/examples/scrypt/pooled.js
+++ b/examples/scrypt/pooled.js
@@ -1,0 +1,37 @@
+'use strict';
+
+const Piscina = require('../../dist/src');
+const { resolve } = require('path');
+const crypto = require('crypto');
+const { promisify } = require('util');
+const randomFill = promisify(crypto.randomFill);
+const { performance, PerformanceObserver } = require('perf_hooks')
+
+const obs = new PerformanceObserver((entries) => {
+  console.log(entries.getEntries()[0].duration)
+});
+obs.observe({ entryTypes: ['measure']});
+
+const piscina = new Piscina({
+  filename: resolve(__dirname, 'scrypt.js'),
+  concurrentTasksPerWorker: 10
+});
+
+async function* generateInput() {
+  let max = parseInt(process.argv[2] || 10);
+  const data = Buffer.allocUnsafe(10);
+  while (max-- > 0) {
+    yield randomFill(data);
+  }
+}
+
+(async function() {
+  performance.mark('start');
+  const keylen = 64;
+
+  for await (const input of generateInput())
+    await piscina.runTask({ input, keylen });
+
+  performance.mark('end');
+  performance.measure('start to end', 'start', 'end');
+})();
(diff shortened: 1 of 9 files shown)


jasnell:
One of the key limitations of this particular example is that it demonstrates that regardless of what we do on the piscina worker thread pooling side, if tasks are scheduling asynchronous work on the libuv threadpool, we're going to be bound by that.. that is, in the pooled example here, we're effectively just scheduling events in one pool that schedules events in another that is being shared, so the performance benefit of using piscina / worker threads is not actually realized.

examples/scrypt/pooled.js
```diff
@@ -0,0 +1,39 @@
...
+  while (max-- > 0) {
+    yield randomFill(data);
+  }
+}
+
+(async function() {
+  performance.mark('start');
+  const keylen = 64;
+  const salt = Buffer.allocUnsafe(16);
+  await randomFill(salt);
```
addaleax, line 32: Not sure how example-y we want this to be, but re-using a salt is not generally a good idea ;)

jasnell: Yeah, I pulled that out to isolate the number of Async tasks happening during the measuring but can definitely move that beck in.

mcollina:
I though libuv was allocating a pool for each loop, or am I wrong?

addaleax:
@mcollina No, the loop is per-process.

jasnell:
Yep. So N worker threads using M libuv threads not M*N. We can see in this example it makes a significant impact

mcollina:
I think it would be important to note this, as the number of libuv threads are lowish.

jasnell:
Running the various examples here on my ubuntu box (8 cores)... with 100 tasks.

image

image

jasnell:
Setting the libuv thread pool size to a larger number doesn't really move the needle too much because we're still bound overall by the performance of the libuv threadpool and we potentially end up in a situation where the worker threads and the libuv threads are contending with each other for CPU. The bottom line lesson here is that there's very little performance gain to be had by offloading async work to a worker thread when that work is already bound to the libuv threadpool.

jasnell:
I think one of the key lessons for this is that the primary performance benefit of workers is going to be moving purely synchronous event loop blocking operations off the main thread (which also should be unsurprising but helps to clarify the specific cases where workers are going to be most beneficial.

PR #14, opened 2020-05-01
tensorflow example

jasnell:
This is an example using tensorflow-node... it is ... interesting and needs further investigation. One issue I'm having with it is after I train the model and save it (`npm run train`) ... when I subsequently load and use the model (`npm run unpooled` or `npm run pooled`) the accuracy of the model is decreased. It likely has something to do with the way tensorflow is being initialized. The code here needs to be cleaned up quite a bit. It's an adaptation of https://www.tensorflow.org/js/tutorials/training/nodejs_training .. specifically the pitch_type.js needs to be adapted to how we're actually using it (saving and loading the model).

As for results... take a look:

image

image

So unpooled is running faster but with order of magnitude higher main event loop blockage.


jasnell:
hold up... found a bug....

jasnell:
Closing this until I figure the bug out. Not only was the streams bit not written correctly but there appears to be a bug running multiple tensorflow node instances in multiple threads concurrently.... investigating.

humphd:
Did this get closed because you think it's not a good approach?  I ask because I'm trying to decide how to deal with the note in the [Tensorflow node.js docs](https://www.tensorflow.org/js/guide/nodejs#production_considerations):

>  If you want to use the Node.js bindings in a production application, like a webserver, you should set up a job queue or set up worker threads so your TensorFlow.js code will not block the main thread.

Reading that made me wonder about using Piscina, but then I saw this and wondered if I should keep looking.  Thanks for any extra info to help me decide.

jasnell:
Really just never got back around to it! I may have time to take a look next week unless you'd like to give it a go.

PR #21, opened 2020-05-01
Add resourceLimits support

jasnell:
Fixes: https://github.com/jasnell/piscina/issues/20

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ This class extends [`EventEmitter`][] from Node.js.
     handling I/O in parallel.
   * `useAtomics`: (`boolean`) Use the [`Atomics`][] API for faster communication
     between threads. This is on by default.
-  * `resourceLimits`: (`object`)
+  * `resourceLimits`: (`object`) See [Node.js new Worker options][]
     * `maxOldGenerationSizeMb`: (`number`) The maximum size of each worker threads
       main heap in MB.
     * `maxYoungGenerationSizeMb`: (`number`) The maximum size of a heap space for
@@ -346,5 +346,6 @@ Piscina development is sponsored by [NearForm Research][].
 [`Atomics`]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics
 [`EventEmitter`]: https://nodejs.org/api/events.html
 [`postMessage`]: https://nodejs.org/api/worker_threads.html#worker_threads_port_postmessage_value_transferlist
+[Node.js new Worker options]: https://nodejs.org/api/worker_threads.html#worker_threads_new_worker_filename_options
 [MIT Licensed]: LICENSE.md
 [NearForm Research]: https://www.nearform.com/research/
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@
 * ✔ Proper async tracking integration
 * ✔ Tracking statistics for run and wait times
 * ✔ Cancelation Support
+* ✔ Supports enforcing memory resource limits
 
 For Node.js 12.x and higher.
 
@@ -147,6 +148,16 @@ This class extends [`EventEmitter`][] from Node.js.
     handling I/O in parallel.
   * `useAtomics`: (`boolean`) Use the [`Atomics`][] API for faster communication
     between threads. This is on by default.
+  * `resourceLimits`: (`object`)
+    * `maxOldGenerationSizeMb`: (`number`) The maximum size of each worker threads
+      main heap in MB.
+    * `maxYoungGenerationSizeMb`: (`number`) The maximum size of a heap space for
+      recently created objects.
+    * `codeRangeSizeMb`: (`number`) The size of a pre-allocated memory range used
+      for generated code.
+
+Use caution when setting resource limits. Setting limits that are too low may
+result in the `Piscina` worker threads being unusable.
 
 ### Method: `runTask(task[, transferList][, filename][, abortSignal])`
 
--- a/examples/resourceLimits/worker.js
+++ b/examples/resourceLimits/worker.js
@@ -3,8 +3,6 @@
 module.exports = () => {
   const array = [];
   while (true) {
-    for (let i = 0; i < 100; i++) {
-      array.push([array]);
-    }
+    array.push([array]);
   }
 };
--- a/examples/resourceLimits/worker.js
+++ b/examples/resourceLimits/worker.js
@@ -1,0 +1,10 @@
+'use strict';
+
+module.exports = () => {
+  const array = [];
+  while (true) {
+    for (let i = 0; i < 100; i++) {
+      array.push([array]);
+    }
+  }
+};
--- a/src/index.ts
+++ b/src/index.ts
@@ -43,6 +43,10 @@ class AbortError extends Error {
   }
 }
 
+type ResourceLimits = Worker extends {
+  resourceLimits? : infer T;
+} ? T : {};
+
 interface Options {
   // Probably also support URL here
   filename? : string | null,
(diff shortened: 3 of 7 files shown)


jasnell:
@addaleax ... not entirely sure why but the resource limits are being rather flaky on macos here.

src/index.ts
```diff
@@ -43,15 +43,24 @@ class AbortError extends Error {
...
   idleTimeout? : number,
   maxQueue? : number,
-  concurrentTasksPerWorker? : number
-  useAtomics? : boolean
+  concurrentTasksPerWorker? : number,
+  useAtomics? : boolean,
+  maxOldGenerationSizeMb? : number,
+  maxYoungGenerationSizeMb? : number,
+  codeRangeSizeMb? : number,
 }
```
addaleax, line 64: If you really do want to list these individually, you might want to use `interface Options extends ResourceLimits` instead

src/index.ts
```diff
@@ -43,15 +43,24 @@ class AbortError extends Error {
   }
 }
 
+interface ResourceLimits {
+  maxOldGenerationSizeMb? : number,
+  maxYoungGenerationSizeMb? : number,
+  codeRangeSizeMb? : number
+}
```
addaleax, line 50:
```suggestion
type ResourceLimits = Worker extends {
  resourceLimits? : infer T;
} ? T : {};
```

(similar to what we’re doing to grab `TransferList` from the official TS definitions)

test/fixtures/resource-limits.js
```diff
@@ -0,0 +1,24 @@
...
+      } else if (space_name === 'old_space') {
+        seenSpaces++;
+      } else if (space_name === 'code_space') {
+        seenSpaces++;
+      }
+    }
+    assert.strictEqual(seenSpaces, 3);
+
+    for (let i = 0; i < 100; i++)
+      array.push([array]);
```
addaleax, line 22:
```suggestion
    array.push([array]);
```

Asserting that the heap spaces are there makes sense for Node.js core but is rather irrelevant here, tbh

README.md
```diff
@@ -147,6 +148,12 @@ This class extends [`EventEmitter`][] from Node.js.
     handling I/O in parallel.
   * `useAtomics`: (`boolean`) Use the [`Atomics`][] API for faster communication
     between threads. This is on by default.
+  * `maxOldGenerationSizeMb`: (`number`) The maximum size of each worker threads
+    main heap in MB.
+  * `maxYoungGenerationSizeMb`: (`number`) The maximum size of a heap space for
+    recently created objects.
+  * `codeRangeSizeMb`: (`number`) The size of a pre-allocated memory range used
+    for generated code.
```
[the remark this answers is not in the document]
jasnell: We can for sure. Although, my TypeScript-foo is weak and I'm not entirely sure how to export the `ResourceLimits` interface without it being breaking since we use `export = Piscina` in index.ts.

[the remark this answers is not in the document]
addaleax: We don’t need to export `ResourceLimits` here at all :)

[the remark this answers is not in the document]
addaleax: Let me know if you’d push that change here, though

addaleax:
> @addaleax ... not entirely sure why but the resource limits are being rather flaky on macos here. Sometimes they run fine, next run they fail inconsistently across versions.
> 
> https://github.com/jasnell/piscina/pull/21/checks?check_run_id=637470841

Hm, odd … I’ll try to take a look at that, thanks for letting me know

README.md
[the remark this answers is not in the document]
jasnell: wait... think I got it figured out that we don't have to.... been way too long since I've mucked around with TypeScript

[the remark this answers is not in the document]
jasnell: Nope, got it figured out :-D

jasnell:
Ok, updated to use the `resourceLimits` object as suggested. The only thing now is that `npm test` is failing consistently now on the new test and really not sure why because the worker is triggering the error.

test/tsconfig.json
```diff
@@ -5,6 +5,6 @@
     "noImplicitAny": false
   },
   "include": [
-    "./**/*.ts"
+    "./**/*.ts", "fixtures/resource-limits.js"
```
addaleax, line 8: Is this necessary? TS doesn’t do anything for `.js` flles, I think

jasnell: what the heck? I think this was vscode trying to be "helpful"

README.md
```diff
@@ -147,6 +148,13 @@ This class extends [`EventEmitter`][] from Node.js.
     handling I/O in parallel.
   * `useAtomics`: (`boolean`) Use the [`Atomics`][] API for faster communication
     between threads. This is on by default.
+  * `resourceLimits`: (`object`)
```
addaleax, line 151: Maybe link to https://nodejs.org/api/worker_threads.html#worker_threads_new_worker_filename_options here? :)

test/fixtures/resource-limits.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+module.exports = () => {
+  const array = [];
+  while (true) {
+    for (let i = 0; i < 100; i++) {
+      array.push([array]);
+    }
```
addaleax, line 8:
```suggestion
    array.push([array]);
```

(The extra loop doesn’t do anything inside `while (true)` :slightly_smiling_face:)

test/test-resourcelimits.ts
```diff
@@ -0,0 +1,24 @@
...
+  const limits : any = worker.options.resourceLimits;
+  is(limits.maxOldGenerationSizeMb, 16);
+  is(limits.maxYoungGenerationSizeMb, 4);
+  is(limits.codeRangeSizeMb, 16);
+  try {
+    await worker.runTask(null);
+  } catch (err) {
+    match(err.message,
+      /Worker terminated due to reaching memory limit: JS heap out of memory/);
+  }
```
addaleax, line 23: I think `rejects` should work here – the issue seems to be that because this is run under `tap`, typescript compilation support is also loaded for the Worker thread, which brings the heap over the limits set here on macOS (but no idea why specifically there)

examples/resourceLimits/worker.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+module.exports = () => {
+  const array = [];
+  while (true) {
+    for (let i = 0; i < 100; i++) {
+      array.push([array]);
+    }
```
addaleax, line 8:
```suggestion
    array.push([array]);
```

jasnell:
On that flaky failure, what it is looking like is that, at least sometimes, the error that's terminating the worker is not getting properly caught and routed to the task Promise.

addaleax:
> On that flaky failure, what it is looking like is that, at least sometimes, the error that's terminating the worker is not getting properly caught and routed to the task Promise.

I don’t think that’s what’s happening – the Promise *is* rejected, but because there was an `'error'` event on the Worker, a new Worker is started, which also OOMs immediately, and then generates an `'error'` event on the `Piscina` instance itself because there is no task associated with that second worker

jasnell:
hmm.. ok, that's fun lol

addaleax:
Right. We could probably try to detect this in some way, but given that rejected Promises and an `'error'` event on the `Piscina` instance are probably the right *results* for this kind of situation, I think we don’t need to. The only thing we could do is add the thread id to the error message to indicate that the errors come from different Worker instances

jasnell:
and for this particular case, attach an `'error'` handler to the Piscina instance...?

addaleax:
@jasnell For the test, I’d just increase the limits. Alternatively, we could also enable forwarding things like `env` and `execArgv` to the `Worker` constructor – setting `execArgv: []` should also fix this (because it removes the TS hook from the list of preloaded modules).

jasnell:
For now, I'll just add the error event handler rather than trying to hunt for limit values that will work consistently. Having the ability to set `env` and `execArgv` on the piscina constructor and having those forward on would be good in a separate PR.

jasnell:
This is good to go 👍

addaleax:
@jasnell There’s https://github.com/jasnell/piscina/pull/21#pullrequestreview-404416829 but if you want you can merge this and I’ll push those changes separately

jasnell:
Yeah, I had just spotted those :-)

PR #26, opened 2020-05-01
Adding support for env, argv, execArgv, and workerData

jasnell:


src/index.ts
```diff
@@ -56,7 +56,11 @@ interface Options {
   maxQueue? : number,
   concurrentTasksPerWorker? : number,
   useAtomics? : boolean,
-  resourceLimits? : ResourceLimits
+  resourceLimits? : ResourceLimits,
+  argv? : string[],
+  execArgv? : string[],
+  env? : any,
```
addaleax, line 62:
```suggestion
  env? : EnvSpecifier,
```

where 

```js
type EnvSpecifier = typeof Worker extends {
  new (filename : string, options?: { env: infer X });
} ? X : never;
```

jasnell:
🎉🥳

PR #27, opened 2020-05-02
Typo fix: Cancelled > Canceled

trivikr:


trivikr:
Actually, cancelation is accurate as per American English https://grammarist.com/spelling/cancel/

PR #34, opened 2020-05-04
Add stream example

jasnell:
@addaleax @mcollina ... here's a super simple example of how streams can be supported.

@addaleax ... interestingly, this example hits an Abort inside Node.js when we replace `createBrotliCompress` with `createGzip` or `createDeflate` within the worker.js...

```
james@ubuntu:~/nearform/piscina/examples/stream$ node index
node[30456]: ../src/node_zlib.cc:323:static void node::{anonymous}::CompressionStream<CompressionContext>::Write(const v8::FunctionCallbackInfo<v8::Value>&) [with bool async = true; CompressionContext = node::{anonymous}::ZlibContext]: Assertion `Buffer::IsWithinBounds(out_off, out_len, Buffer::Length(out_buf))' failed.
 1: 0xa295b0 node::Abort() [node]
 2: 0xa2962e  [node]
 3: 0xae6b1a  [node]
 4: 0xc0251b  [node]
 5: 0xc03ac6  [node]
 6: 0xc04146 v8::internal::Builtin_HandleApiCall(int, unsigned long*, v8::internal::Isolate*) [node]
2
 7: 0x13a5919  [node]
Aborted (core dumped)
```

The reason, for the abort, as far as I can see, is the fact that the `PortDuplex` transfers ownership of the buffer chunk over to the main thread using a transferList, which highlights a significant grey areas with the Streams API ... when using a Writable, and passing a `Buffer`, who takes ownership responsibility for that Buffer? The caller or the receiver? I don't think we have a clear answer for that but we definitely need a fix that does not trigger an Abort.

### DIFF
--- a/examples/stream/port_duplex.js
+++ b/examples/stream/port_duplex.js
@@ -25,7 +25,7 @@ class PortDuplex extends Duplex {
     // Chunk should always be a buffer here
     const temp = new Uint8Array(chunk);
     // TODO(@jasnell): This will need backpressure implemented
-    this.#port.postMessage(temp);
+    this.#port.postMessage(temp, [temp.buffer]);
     callback();
   }
 
--- a/examples/stream/port_duplex.js
+++ b/examples/stream/port_duplex.js
@@ -4,13 +4,11 @@ const { Duplex } = require('stream');
 
 class PortDuplex extends Duplex {
   #port = undefined;
-  #transfer = false;
 
   constructor (port, options) {
     const {
       readable = true,
-      writable = true,
-      transfer = false
+      writable = true
     } = { ...options };
     if (typeof readable !== 'boolean') {
       throw new TypeError('readable must be a boolean');
@@ -18,12 +16,8 @@ class PortDuplex extends Duplex {
     if (typeof writable !== 'boolean') {
       throw new TypeError('writable must be a boolean');
     }
-    if (typeof transfer !== 'boolean') {
-      throw new TypeError('transfer must be a boolean');
-    }
     super({ autoDestroy: true, readable, writable });
     this.#port = port;
-    this.#transfer = transfer;
     this.#port.onmessage = PortDuplex.#onmessage.bind(this);
   }
 
@@ -31,8 +25,16 @@ class PortDuplex extends Duplex {
     if (typeof chunk === 'string') {
       chunk = Buffer.from(chunk, encoding);
     }
-    const transferList = this.#transfer ? [chunk.buffer] : undefined;
-    this.#port.postMessage(chunk, transferList);
+    // Be sure to always copy the chunk here and never use a
+    // transferList. There are several reasons:
+    // a) Buffer instances are most often created off a pool
+    //    and share the same underlying common ArrayBuffer,
+    //    transferring those can break Node.js in many ways.
+    // b) The Buffer instance may still be used by some
+    //    other upstream component. Transferring it here
+    //    will cause unexpected and undefined behavior that
+    //    will likely crash the Node.js process.
+    this.#port.postMessage(chunk);
     callback();
   }
 
--- a/examples/stream/port_duplex.js
+++ b/examples/stream/port_duplex.js
@@ -1,0 +1,67 @@
+'use strict';
+
+const { Duplex } = require('stream');
+
+class PortDuplex extends Duplex {
+  #port = undefined;
+  #transfer = false;
+
+  constructor (port, options) {
+    const {
+      readable = true,
+      writable = true,
+      transfer = false
+    } = { ...options };
+    if (typeof readable !== 'boolean') {
+      throw new TypeError('readable must be a boolean');
+    }
+    if (typeof writable !== 'boolean') {
+      throw new TypeError('writable must be a boolean');
+    }
+    if (typeof transfer !== 'boolean') {
+      throw new TypeError('transfer must be a boolean');
+    }
+    super({ autoDestroy: true, readable, writable });
+    this.#port = port;
+    this.#transfer = transfer;
+    this.#port.onmessage = PortDuplex.#onmessage.bind(this);
+  }
+
+  _write (chunk, encoding, callback) {
+    if (typeof chunk === 'string') {
+      chunk = Buffer.from(chunk, encoding);
+    }
+    const transferList = this.#transfer ? [chunk.buffer] : undefined;
+    this.#port.postMessage(chunk, transferList);
+    callback();
+  }
+
+  _read () {
+    // Do nothing here. A more complete example would
+    // implement proper read/pause behavior.
+  }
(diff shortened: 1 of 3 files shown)


jasnell:
Refs: https://github.com/nodejs/node/issues/33240

jasnell:
I've updated the example to disable the transfer list in the custom duplex by default to avoid the Abort

examples/stream/port_duplex.js
```diff
@@ -0,0 +1,62 @@
...
+
+  _final (callback) {
+    this.#port.postMessage(null);
+    callback();
+  }
+
+  _read () {
+    // Do nothing here. A more complete example would
+    // implement proper read/pause behavior.
+  }
```
addaleax, line 47: Btw, I’ve thought about porting the code from our internal Worker stdio implementation to npm. This is definitely giving more motivation for that :)

addaleax (commented):
The code LGTM but I’m not really sure about this example as it exemplifies several things that are not best practices, including using Workers for file I/O and compression, both of which are already asynchronous operations in Node.js.

If you want a more real-world example, maybe a simple text-processing Worker (e.g. applying a complex regexp to a stream – even just filtering out valid JS identifiers is quite heavy if you want to do it correctly) would be better?

examples/stream/port_duplex.js
jasnell: Likely would be good I think. I'm considering the possibility of Piscina providing a more complete version of `PortDuplex` out of the box.

addaleax: @jasnell The thing is, in the end streaming use cases for Workers are even rarer than use cases for Workers in general. I don’t think we should encourage this pattern if users don’t really need it.

jasnell:
> ... it exemplifies several things that are not best practices, including using Workers for file I/O and compression

Yep, this is actually a lead in to some react server side rendering tests I'm looking at but wanted to make sure the basic idea for the PortDuplex would be sound.

jasnell:
(to be clear, I'm not in a rush to land this particular example, we can keep tweaking it until it's demonstrating the right thing)

addaleax:
@jasnell I’ll try to port over the stream code from core, and I’ll try to look into performance improvements there – theoretically, the same `Atomics` improvements we do for task posting should also make sense for streams, and probably even more so there.

examples/stream/port_duplex.js
jasnell: That's absolutely fair. The key reason I'm thinking through this is to explore the various good and bad ways of using the workers and what the various caveats are.

examples/stream/port_duplex.js
```diff
@@ -0,0 +1,67 @@
...
+    this.#port = port;
+    this.#transfer = transfer;
+    this.#port.onmessage = PortDuplex.#onmessage.bind(this);
+  }
+
+  _write (chunk, encoding, callback) {
+    if (typeof chunk === 'string') {
+      chunk = Buffer.from(chunk, encoding);
+    }
+    const transferList = this.#transfer ? [chunk.buffer] : undefined;
```
mafintosh, line 34: If the buffer was allocated with allocUnsafe this will cause problems due to pooling no?

examples/stream/port_duplex.js
```diff
@@ -4,35 +4,37 @@ const { Duplex } = require('stream');
...
     if (typeof chunk === 'string') {
       chunk = Buffer.from(chunk, encoding);
     }
-    const transferList = this.#transfer ? [chunk.buffer] : undefined;
-    this.#port.postMessage(chunk, transferList);
+    // Be sure to always copy the chunk here and never use a
+    // transferList. There are several reasons:
+    // a) Buffer instances are most often created off a pool
+    //    and share the same underlying common ArrayBuffer,
+    //    transferring those can break Node.js in many ways.
```
addaleax, line 32: See the comment in the Node.js thread, this is true but not necessarily a concern in the future. The sceond concern is definitely there, though :)

examples/stream/port_duplex.js
```diff
@@ -0,0 +1,69 @@
...
+    // Do nothing here. A more complete example would
+    // implement proper read/pause behavior.
+  }
+
+  _destroy (err, callback) {
+    if (err) {
+      // TODO(@jasnell): A more complete example would
+      // handle this error more appropriately.
+      this.#port.close();
+      console.error(err);
```
mcollina, line 51: you should call the callback here.

examples/stream/port_duplex.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+    super({ autoDestroy: true, readable, writable });
+    this.#port = port;
+    this.#port.onmessage = PortDuplex.#onmessage.bind(this);
+  }
+
+  _write (chunk, encoding, callback) {
+    if (typeof chunk === 'string') {
+      chunk = Buffer.from(chunk, encoding);
+    }
```
mcollina, line 27: This will always be a buffer, you are not allowing the encoding to be changed.

examples/stream/port_duplex.js
```diff
@@ -0,0 +1,69 @@
...
+    // transferList. There are several reasons:
+    // a) Buffer instances are most often created off a pool
+    //    and share the same underlying common ArrayBuffer,
+    //    transferring those can break Node.js in many ways.
+    // b) The Buffer instance may still be used by some
+    //    other upstream component. Transferring it here
+    //    will cause unexpected and undefined behavior that
+    //    will likely crash the Node.js process.
+    this.#port.postMessage(chunk);
+    callback();
```
mcollina, line 38: I have a feeling there should be some sort of backpressure built in.

mcollina (commented):
I think this example is complicated enough to either live in its own module or be a part of this one.

In case I would recommend the name https://www.npmjs.com/package/scivolo (water slide).

examples/stream/port_duplex.js
jasnell: Well, there are many ways this is problematic. I'll be updating this such that we always create a new `Uint8Array` copy of the `Buffer` then transferring that new copy. That causes the fewest headaches overall.

jasnell:
> I think this example is complicated enough to either live in its own module or be a part of this one.

I agree. It should definitely live in it's own module and we should be very prescriptive about how and why to use it. @addaleax, I know you were looking at porting the code we use in core to wrap stdio for workers, let's bundle that into the new project. I'll get that repo started by later tonight.

examples/stream/port_duplex.js
jasnell: Absolutely it should

jasnell:
Converted this PR to a draft so it doesn't get landed. Will be iterating on this further.

jasnell:
This is going to be revisited later.

PR #36, opened 2020-05-05
Add coverage for flaky branch

addaleax:
A small percentage of test runs, the Atomics optimization on the
main thread side isn’t covered by the existing tests because the
timing doesn’t work out correctly, rendering our CI flaky.

The added test aims to correct that by coordinating the timing of tasks
in a way that ensures that the optimization is covered.

### DIFF
--- a/test/atomics-optimization.ts
+++ b/test/atomics-optimization.ts
@@ -1,0 +1,66 @@
+import Piscina from '..';
+import { test } from 'tap';
+import { resolve } from 'path';
+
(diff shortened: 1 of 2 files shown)


jasnell:
Well... that's fun.

addaleax:
@jasnell I mean, yeah, it’s a bit silly to have this whole test for a single line of coverage, but flaky CI is even more annoying I would say :woman_shrugging:

PR #42, opened 2020-05-08
Implement queueLoad, queueLoadThreshold, and drain event

jasnell:
@addaleax @mcollina ... this implements a `drain` event that is driven by a `queueLoadThreshold` option.

The `queueLoad` is the ratio of queue size to max queue.

When `queueLoad` falls to equal to or less than `queueLoadThreshold`, the `'drain'` event is emitted.

By default, the `queueLoadThreshold` is `0` meaning that the `'drain'` event fires when the queue is empty.

If specified, the value is between 0 and 1.

There is an example that illustrates the use case (essentially to provide backpressure when the pool is driven by a stream.


mcollina (approved):
lgtm

examples/stream-in/data.csv
```diff
@@ -0,0 +1,5244 @@
+a,b,c,d,e,f,g
```
addaleax, line 1: Add this to the .npmignore?

README.md
```diff
@@ -294,6 +303,12 @@ faster or equal to the given value.
 
 An Array of the `Worker` instances used by this pool.
 
+### Property: `queueLoad` (readonly)
+
+The current ratio of `queueSize` to `maxQueue`. The value is a number between
+`0` and `1`. If `maxQueue` is not specified, the value will always be `1`
+if the `queueSize` is greater than `0`.
```
addaleax, line 310: That doesn’t really make sense, mathematically – if there is no queue load, the ratio would always be `0`?

examples/stream-in/data.csv
jasnell: I'll add a script to generate it and put it in .gitignore and .npmignore

README.md
jasnell: mathematically yeah, but in terms of practical implementation this helps ensure that the drain event always happens when the queue empties. The idea is that if there is no `maxQueue`, having *any* tasks in the queue implies queue load.

src/index.ts
```diff
@@ -632,6 +664,16 @@ class Piscina extends EventEmitter {
     return this.#pool.taskQueue.length;
   }
 
+  get queueLoad () : number {
+    // If maxQueue is Infinity, queueLoad is always either 1 or 0,
+    // otherrwise, queueLoad is the ratio of queueSize / maxQueue
+    if (this.#pool.options.maxQueue === Number.POSITIVE_INFINITY) {
+      return this.queueSize > 0 ? 1 : 0;
+    } else {
+      return this.queueSize / this.#pool.options.maxQueue;
```
addaleax, line 673: `this.#pool.options.maxQueue` can be `0`, btw

README.md
addaleax: @jasnell That’s still a bit inconsistent though … if the ratio is specified as `.5`, then for `maxQueue === 100` it fires at `50`, for `maxQueue === 1000` it fires at `500`, for `maxQueue === 10000` it fires at `5000`, but for `maxQueue === Infinity` it fires at `0`? That seems inconsistent, not just mathematically but also with (my) intuition.

I’m starting to feel like this is mixing a few concerns here … Piscina is a thread pooling library, and this feature is more or less implementing a streams-like backpressure mechanism. I’m not sure what another good solution would be, but it might be easier to fire an event whenever a task is done? That would give more flexibility to the user anyway, and limit the complexity here.

jasnell: Effectively what this is saying is that the `queueLoadThreshold` is only activated when `maxQueue` is not Infinity, because we cannot reasonably calculate a load properly without a finite `maxQueue`... so `maxQueue == Infinity` effectively disables the threshold and causes the `drain` to emit when it would by default (which is when `queueSize` is zero).

src/index.ts
jasnell: Ah, right right. :-)

README.md
addaleax: Well, yeah, the `'drain'` event in this format should be emitted when the `queueSize` drops down to zero, I’m just saying that `queueLoadThreshold` should still be `0` in that case and not `1`? I‘m not disagreeing about the event timing here, to be clear

jasnell: The key challenge I have with firing an event on every completion is that it becomes very chatty. That said, I'm going through today and running a number of experiments on this to see if having the threshold actually yields a measurable/significant performance improvement under various conditions.

For example, assuming the feeding stream is feeding content into the pool at a constant pace, we end up with the following metrics on the main thread:

The runtime duration of the example:
image

The high water mark number of active promises sampled once per event loop turn (as tracked by an async hook):
image

Main thread heap usage sampled once per event loop turn:
image

In this example, when the queue is being fed at a constant rate and the workload of each task is constant, changing the queue load has zero measurable impact. 

The next analysis I want to do is adding variability to the rate data is fed into the queue.. and after that, adding variability to the runtime of each task.

jasnell: Bottom line is I don't plan on landing this as is unless I can demonstrate that the queue threshold makes a measurable impact :-)

jasnell: Ok, results here are a bit more interesting... In this sample run, I introduced a random 0-10 millsecond delay in the feeding stream... I changed up how the graphs are displaying and added an additional metric tracking the wait times (max, 90th %ile, and 99th %ile)...

image

image

image

image

These are all on a 4 core machine, btw...

jasnell: Ok, and the final scenario... maintains the 0-10 ms random latency feeding the stream and adds a 0-100 ms random latency within the workers ... 

image

In all this, duration is measuring the time to completely queue up all the work and get through the stream... not the time to actually complete the work. With the lower maxQueue numbers, the `'drain'` event is firing more frequently and the feeder stream is being paused/resumed more frequently, but overall it's not having a drastic impact on feed time no matter how often it fires, which is very good to see.

image

image

(the heap on the last one is a blip... running that 100-0.0 test again the heap use falls in line with the others)

image

So it's definitely looking like load threshold isn't going to be significant to justify the additional complexity here but this has still been a valuable exercise in exploring the performance of the pool.

jasnell:
Ok, given the analysis on this, @addaleax was right from the beginning 😆 ... but this was still a good exercise to go through as it validated a few of the other performance tuning bits we already have in place. Specifically: 

`maxQueue` and `maxThreads` are the two key performance knobs we can turn here assuming the pool is being actively fed. Based on all the runs that I made, there's no benefit to `maxQueue` being set to more than a modest multiple of `maxThreads` (e.g. for `maxThreads = 4`, a `maxQueue` of `8` or `16` is reasonable under the assumption that feeding the queue is less expensive an operation than processing a queued item. Any value larger than that just blows out the wait times without any appreciable benefit whatsoever.

I'm going to close this and open a separate PR that emits `'drain'` whenever the queue size hits `0`

PR #43, opened 2020-05-08
Implement drain event

jasnell:
Simplified alternative to https://github.com/jasnell/piscina/pull/42 ... only implements the drain event


src/index.ts
```diff
@@ -519,9 +520,16 @@ class ThreadPool {
     this.waitTime.recordValue(now - taskInfo.created);
     taskInfo.started = now;
     workerInfo.postTask(taskInfo);
+    this._maybeDrain();
     return ret;
   }
 
+  _maybeDrain () {
+    if (this.taskQueue.length === 0) {
+      this.publicInterface.emit('drain');
```
addaleax, line 529: I think it it would make sense to turn `ThreadPool` into an `AsyncResource` subclass as well, to make async tracking for these events a bit cleaner. That’s probably already an issue with `'error'` as well, so I’ll look into that after merging.

PR #51, opened 2020-05-11
Add n-api example

jasnell:
Uses a simple native addon as the worker implementation

### DIFF
--- a/examples/n-api/binding.gyp
+++ b/examples/n-api/binding.gyp
@@ -1,0 +1,12 @@
+{
+  "targets": [{
+    "target_name": "example",
+    "include_dirs": [
+      "<!(node -e \"require('napi-macros')\")",
+      "deps/ngtcp2/includes"
+    ],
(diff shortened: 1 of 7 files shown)


examples/n-api/binding.gyp
```diff
@@ -0,0 +1,12 @@
+{
+  "targets": [{
+    "target_name": "example",
+    "include_dirs": [
+      "<!(node -e \"require('napi-macros')\")",
```
addaleax, line 5: I don’t think this is being used in the `.cc` file, and I wouldn’t recommend using it for examples either. It only makes sense if you need to write C code and C++ is not an option for you, for C++ https://github.com/nodejs/node-addon-api is much, much nicer.

jasnell: ah, right. meant to pull that back out since I wasn't actually using it

jasnell:
Fixed!

PR #57, opened 2020-05-14
Wait for Workers to be ready before considering them as task candidates

addaleax:
Refactor the Worker pool management in order to better accommodate the
async nature of how they are started, and use that to wait until the
target module has been loaded before considering the Worker as a
task processing candidate in order to improve performance.

This also includes bugfixes for working around a crash in Node.js core.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -10,7 +10,7 @@ import assert from 'assert';
 import { Histogram, build } from 'hdr-histogram-js';
 import { performance } from 'perf_hooks';
 import hdrobj from 'hdr-histogram-percentiles-obj';
-import { RequestMessage, ResponseMessage, StartupMessage, commonState, kResponseCountField, kRequestCountField, kFieldCount } from './common';
+import { ReadyMessage, RequestMessage, ResponseMessage, StartupMessage, commonState, kResponseCountField, kRequestCountField, kFieldCount } from './common';
 import { version } from '../package.json';
 
 const cpuCount : number = (() => {
@@ -69,7 +69,7 @@ interface FilledOptions extends Options {
   minThreads : number,
   maxThreads : number,
   idleTimeout : number,
-  maxQueue : number | 'auto',
+  maxQueue : number,
   concurrentTasksPerWorker : number,
   useAtomics: boolean
 }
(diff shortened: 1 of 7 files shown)


src/index.ts
```diff
@@ -69,7 +69,7 @@ interface FilledOptions extends Options {
   minThreads : number,
   maxThreads : number,
   idleTimeout : number,
-  maxQueue : number | 'auto',
+  maxQueue : number,
```
jasnell, line 72: Is this intentional?

src/index.ts
```diff
@@ -255,29 +361,35 @@ class WorkerInfo {
...
     this.publicInterface = publicInterface;
-    this.workers = [];
     this.taskQueue = [];
     this.runTime = build({ lowestDiscernibleValue: 1 });
     this.waitTime = build({ lowestDiscernibleValue: 1 });
 
     const filename =
       options.filename ? maybeFileURLToPath(options.filename) : null;
-    this.options = { ...kDefaultOptions, ...options, filename };
+    this.options = { ...kDefaultOptions, ...options, filename, maxQueue: 0 };
```
jasnell, line 392: Is forcing maxQueue to 0 here intentional?

test/idle-timeout.ts
```diff
@@ -36,8 +36,8 @@ test('idle timeout will let go of threads early', async ({ is }) => {
   is(pool.threads.length, 1);
 
   const secondTasks = [
-    pool.runTask(busyLoop(100) + getThreadId),
-    pool.runTask(busyLoop(100) + getThreadId)
+    pool.runTask(busyLoop(500) + getThreadId),
+    pool.runTask(busyLoop(500) + getThreadId)
```
jasnell, line 40: Saw this previously but hadn't asked... would a sync sleep using Atomics.wait() work here also? Or does it need to busy loop the cpu?

src/index.ts
addaleax: Yes, because `maxQueue` is never `'auto'` there. In particular, this is relevant specifically when you’re using `maxQueue: 'auto'`, because then you have to check the value of `pool.options.maxQueue`, and now when you’re using TS you don’t have to explicitly assert that that is a number anymore.

src/index.ts
addaleax: Yeah, it’s to make sure that `'auto'` doesn’t end up in `this.options` at any point, to comply with the changed type signature

test/idle-timeout.ts
addaleax: Yeah, I think `Atomics.wait` would work too. I’ll try to refactor this test anyway since it still seems to be failing in CI

addaleax:
CI is failing only on Windows and only on one test, but I couldn’t reproduce that locally on Windows, so I’ll apologize in advance for the notifications you’ll get from me debugging this against CI :see_no_evil:

addaleax:
Okay, I've managed to reproduce this after figuring out that it specifically happens when the CPU count is 2 (which translates to manually setting CPU affinity in the Task Manager locally 😬) I think for further debugging I'll need to compile a fresh Node.js on Windows, which means that I'll let my laptop do that for the next hour or so and give up for today :smile:

tuananh:
@addaleax can we let the worker define the `ready` state?

use case: initialize webassembly module?

jasnell:
@tuananh ... With the changes @addaleax is making here that should be feasible. What I've considered is allowing the worker to register an init function during it's bootstrap that piscina would await before it reports itself ready.

PR #59, opened 2020-05-15
Implement Piscina.move

jasnell:
Example worker.js

```js
const { move } = require('piscina');

module.exports = () => {
  // transfers the arraybuffer rather than cloning it.
  return move(new ArrayBuffer(10));
};
```

Fixes: https://github.com/jasnell/piscina/issues/58

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,6 +424,79 @@ Is `true` if this code runs inside a `Piscina` threadpool as a Worker.
 
 Provides the current version of this library as a semver string.
 
+### Static method: `move(value)`
+
+By default, any value returned by a worker function will be cloned when
+returned back to the Piscina pool, even if that object is capable of
+being transfered. The `Piscina.move()` method can be used to wrap and
+mark transferable values such that they will by transfered rather than
+cloned.
+
+The `value` may be any object supported by Node.js to be transferable
+(e.g. `ArrayBuffer`, any `TypedArray`, or `MessagePort`), or any object
+implementing the `Transferable` interface.
+
+```js
+const { move } = require('piscina');
+
+module.exports = () => {
+  return move(new ArrayBuffer(10));
+}
+```
+
+The `move()` method will throw if the `value` is not transferable.
+
+The object returned by the `move()` method should not be set as a
+nested value in an object. If it is used, the `move()` object itself
+will be cloned as opposed to transfering the object it wraps.
+
+#### Interface: `Transferable`
+
+Objects may implement the `Transferable` interface to create their own
+custom transferable objects. This is useful when an object being
+passed into or from a worker contains a deeply nested transferable
+object such as an `ArrayBuffer` or `MessagePort`.
+
+`Transferable` objects expose two properties inspected by Piscina
+to determine how to transfer the object. These properties are
+named using the special static `Piscina.transferableSymbol` and
+`Piscina.valueSymbol` properties:
+
+* The `Piscina.transferableSymbol` property provides the object
+  (or objects) that are to be included in the `transferList`.
+
+* The `Piscina.valueSymbol` property provides a surrogate value
+  to transmit in place of the `Transferable` itself.
+
+Both properties are required.
+
+For example,
+
+```js
+const {
+  move,
+  transferableSymbol,
+  valueSymbol
+} = require('piscina');
+
+module.exports = () => {
+  const obj = {
+    a: { b: new Uint8Array(5); },
+    c: { new Uint8Array(10); },
+
+    get [transferableSymbol]() {
+      // Transfer the two underlying ArrayBuffers
+      return [this.a.b.buffer, this.c.buffer];
+    }
+
+    get [valueSymbol]() {
+      return { a: { b: this.b }, c: this.c };
+    }
+  };
+  return move(obj);
+};
+```
+
 ## Current Limitations (Things we're working on / would love help with)
 
 * Improved Documentation
--- a/src/common.ts
+++ b/src/common.ts
@@ -28,6 +28,40 @@ export const commonState = {
   workerData: undefined
 };
 
+// Internal symbol used to mark Transferable objects returned
+// by the Piscina.move() function
+const kMovable = Symbol('Piscina.kMovable');
+export const kTransferable = Symbol.for('Piscina.transferable');
+export const kValue = Symbol.for('Piscina.valueOf');
+
+// True if the object implements the Transferable interface
+export function isTransferable (value : any) : boolean {
+  return value != null &&
+         typeof value === 'object' &&
+         kTransferable in value &&
+         kValue in value;
+}
+
+// True if object implements Transferable and has been returned
+// by the Piscina.move() function
+export function isMovable (value : any) : boolean {
+  return isTransferable(value) && value[kMovable] === true;
+}
+
+export function markMovable (value : object) : void {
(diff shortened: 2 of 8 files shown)


addaleax:
Just fyi, I am reviewing this and I’ll have a few comments ;)

jasnell:
> I’ll have a few comments ;)

I suspected as much 😄

README.md
```diff
@@ -424,6 +424,70 @@ Is `true` if this code runs inside a `Piscina` threadpool as a Worker.
...
+passed into or from a worker contains a deeply nested transferable
+object such as an `ArrayBuffer` or `MessagePort`.
+
+A `Transferable` object is any object that exposes a public `transferable`
+property that provides an array of values that should be listed in the
+`transferList` that Piscina uses to communicate across threads.
+
+`Transferable` objects *should* also implement `valueOf()` function that
+returns a surrogate value to transmit in place of the `Transferable`
+itself.
```
addaleax, line 466: Can we use Symbols for `transferable` and `valueOf`? That’s more or less exactly what Symbols are there for, and they would avoid conflicts with existing properties of that name (and, for examples, the semantics of `valueOf()` here are closer to those of `toJSON()`, but both already have specific semantics in JS itself)

src/common.ts
```diff
@@ -28,6 +28,33 @@ export const commonState = {
...
+// by the Piscina.move() function
+export function isMovable (value : any) : boolean {
+  return isTransferable(value) && value[kMovable] === true;
+}
+
+export function markMovable (value: any) : void {
+  Object.defineProperty(value, kMovable, {
+    enumerable: false,
+    value: true
+  });
```
addaleax, line 50: Is `enumerable` listed here in order to be explicit, as opposed to `writable`/`configurable`? I don’t think `configurable: true, writable: true` would be a bad choice here

src/common.ts
```diff
@@ -28,6 +28,33 @@ export const commonState = {
...
+// by the Piscina.move() function
+const kMovable = Symbol('Piscina.kMovable');
+
+// True if object implements Transferable and has been returned
+// by the Piscina.move() function
+export function isMovable (value : any) : boolean {
+  return isTransferable(value) && value[kMovable] === true;
+}
+
+export function markMovable (value: any) : void {
```
addaleax, line 46:
```suggestion
export function markMovable (value : object) : void {
```

src/index.ts
```diff
@@ -860,6 +908,22 @@ class Piscina extends EventEmitterAsyncResource {
...
+      if (types.isAnyArrayBuffer(val)) {
+        val = new DirectlyTransferable(val);
+      } else if ((types as any).isArrayBufferView(val)) {
+        val = new ArrayBufferViewTransferable(val as ArrayBufferView);
+      } else if (val instanceof MessagePort) {
+        val = new DirectlyTransferable(val);
+      } else {
+        throw new TypeError('value is not transferable');
+      }
+    }
```
addaleax, line 923: Fwiw, `instanceof MessagePort` is an incomplete brand check (because it’s perfectly fine for ports to live in different Contexts), and generally there will not be an exhaustive list of types to check against – for example, we are adding support for [`KeyObject`](https://github.com/nodejs/node/pull/33360) right now.

That’s tricky and I’m not really sure how to work around that.

src/common.ts
```diff
@@ -28,6 +28,33 @@ export const commonState = {
...
+  Object.defineProperty(value, kMovable, {
+    enumerable: false,
+    value: true
+  });
+}
+
+export interface Transferable {
+  readonly transferable : any;
+  valueOf() : any;
```
addaleax, line 55: `transferable` and the return value of `valueOf()` should be `object`s, right?

README.md
jasnell: Yeah, I was debating that also. Symbols are a bit less ergonomic I think but definitely doable. Will switch these to use `Symbol.for('Piscina...')` symbols...

src/common.ts
jasnell: yeah, just to be explicit. And yeah, configurable/writable are fine also

src/index.ts
addaleax: I would probably just allow any kind of object here, and if it’s not actually transferable, then so be it.

We can still improve the experience for TypeScript users here if we define this as 

```js
static move (val : Transferable | TransferListItem | ArrayBufferView) {
  ...
```

where 

```js
type TransferListItem = TransferList extends (infer T)[] ? T : never;
```

jasnell: Yeah, I'm not happy with this one either. What would be interesting is a `isTransferable()` in *core* that matches whatever set of transferables that version of Node.js supports (e.g. I'm not certain the `KeyObject` one will be backported any time soon).

Btw, is the `KeyObject` one actually going to allow the key to be transferred or just cloned efficiently?

addaleax: Right, core is more or less the only authoritative source on whether something can be transferred or not.

The `KeyObject` one supports cloning in its current state, but it could also easily be extended to transferring.

README.md
addaleax: We can still export those symbols on the `Piscina` class itself, that should be fine in terms of ergonomics I think

src/index.ts
jasnell: And yeah, relaxing it to any object is likely best for now.

src/index.ts
```diff
@@ -860,6 +908,22 @@ class Piscina extends EventEmitterAsyncResource {
   static get Piscina () {
     return Piscina;
   }
+
+  static move (val : any) {
+    if (!isTransferable(val)) {
+      if (types.isAnyArrayBuffer(val)) {
+        val = new DirectlyTransferable(val);
+      } else if ((types as any).isArrayBufferView(val)) {
```
addaleax, line 916: Opened https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44769 to fix this, btw

jasnell:
@addaleax ... updated! :-)

src/worker.ts
```diff
@@ -112,12 +123,17 @@ function onMessage (
 
   (async function () {
     let response : ResponseMessage;
+    const transferList : any[] = [];
```
addaleax, line 126:
```suggestion
    const transferList : object[] = [];
```

src/index.ts
```diff
@@ -84,6 +101,28 @@ const kDefaultOptions : FilledOptions = {
...
+
+  get [kTransferable] () : object { return this.#value; }
+
+  get [kValue] () : object { return this.#value; }
+}
+
+class ArrayBufferViewTransferable implements Transferable {
+  #view : ArrayBufferView;
+  constructor (view : ArrayBufferView) {
+    this.#view = view;
```
addaleax, line 118: Should we verify that `view.buffer.byteLength === view.byteLength` here, to make this less easy to get wrong?

(Aside, since these classes are internal and not exposed to users, I don’t think you really need to make the properties private – currently, that still transpiles to a WeakMap implementation)

jasnell: I don't think so because there are cases (albeit edge cases for sure) where that check doesn't work. I'd say let's leave this and maybe reference the warning we just added to the core docs.

And yeah, these properties don't need to be private

addaleax: Hm, okay… I know there are edge cases, but those seem kinda fabricated to me ;)

PR #60, opened 2020-05-15
Handle errors in stream example

jasnell:


examples/stream-in/progress.js
```diff
@@ -0,0 +1,39 @@
+const { EventEmitter } = require('events');
```
addaleax, line 1:
```suggestion
'use strict';
const { EventEmitter } = require('events');
```

since this is plain JS?

PR #65, opened 2020-05-16
Implement custom task queues

jasnell:
Include example of implementing a priority queue

Refs: https://github.com/jasnell/piscina/issues/48

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -20,6 +20,10 @@ import {
   kRequestCountField,
   kFieldCount,
   Transferable,
+  Task,
+  TaskQueue,
+  kQueueOptions,
+  isTaskQueue,
   isTransferable,
   markMovable,
   isMovable,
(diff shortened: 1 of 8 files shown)


src/index.ts
```diff
@@ -190,6 +217,10 @@ class TaskInfo extends AsyncResource {
     this.runInAsyncScope(this.callback, null, err, result);
     this.emitDestroy(); // `TaskInfo`s are used only once.
   }
+
+  get [kQueueOptions] () : any | null {
```
addaleax, line 221: The docs say this is `object | null` – this should either be `object | null`, or if not, then `any`, because `any | null` is the same as `any` :)

jasnell:
@mcollina ... I was thinking about a separate `piscina-priority-queue` module that would provide an implementation... e.g.

```js
const Piscina = require('piscina');
const PiscinaPriorityQueue = require('piscina-priority-queue');

const p = new Piscina({
   taskQueue: new PiscinaPriorityQueue()
});
```

PR #67, opened 2020-05-17
Support async loaded workers

jasnell:
@addaleax ... this works but I'm not convinced yet that it's the best way. Would definitely appreciate a look over.

Fixes https://github.com/jasnell/piscina/issues/63

### DIFF
--- a/examples/async_load/worker.js
+++ b/examples/async_load/worker.js
@@ -1,0 +1,9 @@
+'use strict';
+
+const { promisify } = require('util');
+const sleep = promisify(setTimeout);
+
+module.exports = (async () => {
+  await sleep(500);
+  return () => console.log('hello from an async loaded CommonJS worker');
+})();
--- a/src/worker.ts
+++ b/src/worker.ts
@@ -36,7 +36,7 @@ async function getHandler (filename : string) : Promise<Function | null> {
   try {
     handler = await importESM(pathToFileURL(filename).href);
     if (typeof handler !== 'function') {
-      handler = (handler as any).default;
+      handler = await (handler as any).default;
     }
   } catch {}
   if (typeof handler !== 'function') {
@@ -44,7 +44,7 @@ async function getHandler (filename : string) : Promise<Function | null> {
     // `require(filename)`.
     handler = await import(filename);
     if (typeof handler !== 'function') {
-      handler = (handler as any).default;
+      handler = await (handler as any).default;
     }
   }
   if (typeof handler !== 'function') {
--- a/README.md
+++ b/README.md
@@ -138,6 +138,29 @@ const piscina = new Piscina({
 })();
 ```
 
+### Delaying Availability of Workers
+
+A worker thread will not be made available to process tasks until Piscina
+determines that it is "ready". By default, a worker is ready as soon as
+Piscina loads it and acquires a reference to the exported handler function.
+
+There may be times when the availability of a worker may need to be delayed
+longer while the worker initializes any resources it may need to operate.
+To support this case, the worker module may export a `Promise` that resolves
+the handler function as opposed to exporting the function directly:
+
+```js
+async function initialize() {
+  await someAsyncInitializationActivity();
+  return ({ a, b }) => a + b;
+}
+
+module.exports = initialize();
+```
+
+Piscina will await the resolution of the exported Promise before marking
+the worker thread available.
+
 ### Backpressure
 
 When the `maxQueue` option is set, once the `Piscina` queue is full, no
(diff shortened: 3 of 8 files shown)


examples/async_load/worker.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+const { promisify } = require('util');
+const sleep = promisify(setTimeout);
+
+module.exports = (async () => {
+  await sleep(500);
+  return () => process._rawDebug('hello from an async loaded CommonJS worker');
```
jasnell, line 8: @addaleax ... there's definitely an issue with using `console.log()` and `process.stdout`/`process.stderr` in workers. Specifically, it'll work once or twice in a worker then on subsequent jobs it will just stop. You *should* be able to see that here if you change the `process._rawDebug()` statements to `console.log()` in both workers, then run the example over and over... You'll see something like...

```
james@ubuntu:~/nearform/piscina/examples/async_load$ node index
hello from an async loaded CommonJS worker
hello from an async loaded ESM worker
james@ubuntu:~/nearform/piscina/examples/async_load$ node index
hello from an async loaded ESM worker
james@ubuntu:~/nearform/piscina/examples/async_load$ node index
hello from an async loaded ESM worker
james@ubuntu:~/nearform/piscina/examples/async_load$ node index
hello from an async loaded ESM worker
james@ubuntu:~/nearform/piscina/examples/async_load$ node index
hello from an async loaded ESM worker
james@ubuntu:~/nearform/piscina/examples/async_load$ node index
hello from an async loaded ESM worker
james@ubuntu:~/nearform/piscina/examples/async_load$ node index
hello from an async loaded ESM worker
hello from an async loaded CommonJS worker
james@ubuntu:~/nearform/piscina/examples/async_load$ node index
hello from an async loaded CommonJS worker
hello from an async loaded ESM worker
james@ubuntu:~/nearform/piscina/examples/async_load$ node index
hello from an async loaded ESM worker
james@ubuntu:~/nearform/piscina/examples/async_load$ node index
hello from an async loaded ESM worker
hello from an async loaded CommonJS worker
```

I'm only using `process._rawDebug()` here because of the issue.

src/worker.ts
```diff
@@ -36,15 +36,15 @@ async function getHandler (filename : string) : Promise<Function | null> {
...
     if (typeof handler !== 'function') {
-      handler = (handler as any).default;
+      handler = await (handler as any).default;
     }
   } catch {}
   if (typeof handler !== 'function') {
     // With our current set of TypeScript options, this is transpiled to
     // `require(filename)`.
-    handler = await import(filename);
+    handler = await (await import(filename));
```
jasnell, line 45: the double await here is... interesting... not 100% sure we need it ... going to keep testing

jasnell: Yep, looks like it's not needed... ;-)

examples/async_load/worker.js
addaleax: @jasnell I can’t reproduce that (x64 Ubuntu 18.04 with only `process._rawDebug()` → `console.log()`) :/

jasnell: Super odd. I'll come up with a pure node repo and open an issue

addaleax: @jasnell Does setting `useAtomics: false` as a Piscina option make a difference for you?

jasnell: Yeah, setting `useAtomics: false` *appears* to make it work correctly. Also, adding some random delay into each handler also makes it work. Possibly a race condition when multiple workers running at the same time console.log/process.stdout.write at nearly the same time?

addaleax: @jasnell I think it’s the fact that `process.stdout.write()` is an asynchronous operation whose livetime can extend beyond the one of the task itself – the `Atomics.wait()` call in the Worker doesn’t know that and will enter waiting state while more writes might be waiting behind it.

Adjusting the `while (currentTasks === 0) {` condition in `atomicsWaitLoop()` to account for it by adding `&& !process.stdout._writableState.writing && !process.stderr._writableState.writing` might help, but I don’t really want to try that without a reproduction

addaleax: https://github.com/jasnell/piscina/pull/68 should fix this :)

jasnell:
Awesome, #68 fixed the console.log issue it appears. What are your thoughts on this actual PR, @addaleax?

jasnell:
I'm going to add a few comments to the docs on using this before landing.

addaleax:
The README additions also sound good :+1:

PR #69, opened 2020-05-18
Account for loadavg when queuing tasks

### LINKED ISSUE #47: Feature: Take os.loadavg() into account
It would be nice if `piscina` took `os.loadavg()` into account when deciding how many workers to use in a given situation, to dynamically respond to the system load. This can be helpful considering that the application that uses `piscina` may not be the main application running on a given host, and that the up to `1.5 × number of CPUs` `maxThreads` default may be too much in a lot of circumstances.

The idea here is that we would not post new tasks until either less than `minThreads` tasks are currently running or the system drops below x CPU load (where x could be e.g. 100 % by default, but configurable).

In a similar vein, it would be nice to be able to set scheduling paramters for the spawned threads, but that would require upstream support in libuv first, then in Node.js, then here.

jasnell:
@addaleax ... please take a look to see if this approach is acceptable.

This adds a new _acceptableCpuLoad() check that is currently only used when submitting a new task. If the function returns false, the task is queued, even if it otherwise would be dispatched. The check is not applied when a worker comes available with the idea that it is always acceptable for a worker to take on a new task, regardless of cpu load, if there are tasks available.

The check is disabled by default (the default `cpuLoadAvgThreshold` is set to `Infinity`), and disabled entirely on Windows since the loadavg there is not calculated.

The threshold must be a number between 0.0 and 1.0, although we might want to remove the upper limit and allow any value to be specified. The reason is because the cpu loadavg could easily regularly exceed the capacity of available CPU cores.

The load average is calculated by taking the 1-min load average and dividing by the number of CPU cores available (e.g. `os.loadavg()[0] / cpus().length` in order to normalize the values. Alternatively, we could eliminate the upper limit on `cpuLoadAvgThreshold` and avoid the division but then users would need to set a threshold that takes the number of CPUs into consideration.

(Marked this as a draft until we're sure this is the way we want to do it)

Fixes: https://github.com/jasnell/piscina/issues/47

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -454,6 +454,21 @@ class WorkerInfo extends AsynchronouslyCreatedResource {
   }
 }
 
+function waitForAcceptableCpuLoad (pool : ThreadPool) {
+  // TODO(@jasnell): Not entirely clea how to force this in test.
+  pool.loadTimer = setTimeout(/* istanbul ignore next */ () => {
+    if (pool._acceptableCpuLoad()) {
+      const workerInfo : WorkerInfo | null = pool.workers.findAvailable();
+      if (workerInfo) {
+        pool._dispatchTask(workerInfo);
+      }
+    } else {
+      // Keep waiting ...
+      waitForAcceptableCpuLoad(pool);
+    }
+  }, 100);
+}
+
 class ThreadPool {
   publicInterface : Piscina;
   workers : AsynchronouslyCreatedResourcePool<WorkerInfo>;
@@ -467,6 +482,7 @@ class ThreadPool {
   inProcessPendingMessages : boolean = false;
   startingUp : boolean = false;
   workerFailsDuringBootstrap : boolean = false;
+  loadTimer? : NodeJS.Timeout;
 
   constructor (publicInterface : Piscina, options : Options) {
     this.publicInterface = publicInterface;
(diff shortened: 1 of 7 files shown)


src/index.ts
```diff
@@ -629,7 +632,22 @@ class ThreadPool {
...
+      return true;
+    }
+    return loadavg()[0] / cpuCount <= this.options.cpuLoadAvgThreshold;
+  }
+
   _onWorkerAvailable (workerInfo : WorkerInfo) : void {
+    // We do not check _acceptableCpuLoad here with the idea
+    // that if a worker is coming available, then it should
+    // always be permitted to take on a new task if one is
+    // available.
```
addaleax, line 650: I’m not sure whether it makes sense to distinguish this case from the one below?

src/index.ts
```diff
@@ -702,9 +720,10 @@ class ThreadPool {
       });
     }
 
-    // If there is a task queue, there's no point in looking for an available
-    // Worker thread. Add this task to the queue, if possible.
-    if (this.taskQueue.size > 0) {
+    // If there is a task queue, or there is unacceptable CPU load,
+    // there's no point in looking for an available Worker thread.
+    // Add this task to the queue, if possible.
+    if (this.taskQueue.size > 0 || !this._acceptableCpuLoad()) {
```
addaleax, line 726: If we hit this condition, I think should probably start a timer that checks whether CPU load has become acceptable or not, otherwise the task will hang around indefinitely.

jasnell: Yeah, that's where I was leaning but wanted to see if you ended up in the same place ;-) This is exactly why I skipped the check on the worker available event above.

jasnell:
Going to separate out the first commit into a separate PR now.

jasnell:
Closing this as I think we're going a different direction with it.

PR #73, opened 2020-05-22
Add niceIncrement option on Linux

addaleax:
The OS restriction is due to the fact that macOS does not provide
per-thread priority values.

### DIFF
--- a/test/nice.ts
+++ b/test/nice.ts
@@ -1,0 +1,31 @@
+import Piscina from '..';
+import { resolve } from 'path';
+import { test } from 'tap';
+
+test('can set niceness for threads on Linux', {
+  skip: process.platform !== 'linux'
+}, async ({ is }) => {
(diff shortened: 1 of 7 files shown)


mcollina:
nice!

test/nice.ts
```diff
@@ -0,0 +1,31 @@
...
+  // @ts-ignore
+  const currentNiceness = (await import('nice-napi')).default(0);
+  const result = await worker.runTask('require("nice-napi")()');
+
+  // niceness is capped to 19 on Linux.
+  const expected = Math.min(currentNiceness + 5, 19);
+  is(result, expected);
+});
+
+test('setting niceness never odes anything bad', async ({ is }) => {
```
jasnell, line 23:
```suggestion
test('setting niceness never does anything bad', async ({ is }) => {
```

PR #77, opened 2020-05-26
Use once listener for abort event

jasnell:
There's no reason to keep the listener around.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -53,7 +53,6 @@ interface AbortSignalEventTarget {
   aborted? : boolean;
 }
 interface AbortSignalEventEmitter {
-  on : (name : 'abort', listener : () => void) => void;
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -41,19 +41,27 @@ const cpuCount : number = (() => {
...
 interface AbortSignalEventTarget {
-  addEventListener : (name : 'abort', listener : () => void) => void;
+  addEventListener : (
+    name : 'abort',
+    listener : () => void,
+    options? : AbortSignalEventTargetAddOptions) => void;
   aborted? : boolean;
 }
 interface AbortSignalEventEmitter {
   on : (name : 'abort', listener : () => void) => void;
```
addaleax, line 56:
```suggestion
```

PR #86, opened 2020-07-05
Prefer loading as CJS before attempting ESM

### LINKED ISSUE #85: Warnings about experimental ESM module loader while CommonJS is used
Hi,

I was trying to use Piscina in production and I've used CommonJS for all my code, including worker.

Basically, my worker looks like this:
```javascript
async function initialize() {
    return (action) => {
         // do something cpu-intensive here, based on action received
    }
}
module.exports = initialize();
```

Each time the worker starts, I am getting a message in the console:
```
(node:94710) ExperimentalWarning: The ESM module loader is experimental.
```

That unfortunate, because it is polluting logs and is not in a json format which is the format of my app logger (I use Fastify with Pino).

Is it possible to check is file is written in CommonJS format and use `require` instead of `import` to load worker code?
Or maybe use `require` when file extension is `cjs` (taking into account package type if any?) ?

addaleax:
Use the CJS loader first and only load using ESM if that hasn’t worked.

This works around a feature warning for ESM on Node v12.x.

Fixes: https://github.com/piscinajs/piscina/issues/85

### DIFF
--- a/src/worker.ts
+++ b/src/worker.ts
@@ -21,9 +21,16 @@ let useAtomics : boolean = true;
 
 // Get `import(x)` as a function that isn't transpiled to `require(x)` by
 // TypeScript for dual ESM/CJS support.
-const importESM : (specifier : string) => Promise<any> =
-  // eslint-disable-next-line no-eval
-  eval('(specifier) => import(specifier)');
+// Load this lazily, so that there is no warning about the ESM loader being
+// experimental (on Node v12.x) until we actually try to use it.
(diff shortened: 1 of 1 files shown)


jasnell (approved):
@addaleax ... are there any semver-major ramifications on this?

addaleax:
@jasnell I think it might technically lead to different results if `"exports"` in `package.json` contains different implementations for `"require"` and `"import"`. However, that is rare in practice, and would be considered bad practice anyway.

PR #94, opened 2020-11-04
Update dependencies, add trackUnmanagedFds

jasnell:
Ping @addaleax ... can you take a look at the typescript type fixes in this to see if they make sense.

This will be a semver-major update given the dependency updates.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,21 +32,22 @@
   ],
   "license": "MIT",
   "devDependencies": {
-    "@types/node": "^14.0.19",
-    "@typescript-eslint/eslint-plugin": "^3.6.0",
-    "@typescript-eslint/parser": "^3.6.0",
+    "@types/node": "^14.14.6",
+    "@typescript-eslint/eslint-plugin": "^4.6.1",
+    "@typescript-eslint/parser": "^4.6.1",
     "abort-controller": "^3.0.0",
     "concat-stream": "^2.0.0",
-    "gen-esm-wrapper": "^1.0.6",
-    "snazzy": "^8.0.0",
-    "standardx": "^5.0.0",
-    "tap": "^14.10.7",
-    "typescript": "^3.9.6"
+    "gen-esm-wrapper": "^1.1.0",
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -981,7 +993,7 @@ class Piscina extends EventEmitterAsyncResource {
     return Piscina;
   }
 
-  static move (val : Transferable | TransferListItem | ArrayBufferView) {
+  static move (val : Transferable | TransferListItem | ArrayBufferView | ArrayBuffer | MessagePort) {
```
addaleax, line 996: These two should already have been captured through `TransferListItem`, I think?

addaleax (dismissed):

src/index.ts
jasnell: I thought they would too but they weren't in local testing. Not sure why exactly

mcollina (dismissed):
lgtm

package.json
```diff
@@ -32,21 +32,22 @@
...
+    "tap": "^14.10.8",
+    "typescript": "^4.0.5"
   },
   "dependencies": {
     "eventemitter-asyncresource": "^1.0.0",
-    "hdr-histogram-js": "^1.2.0",
-    "hdr-histogram-percentiles-obj": "^2.0.1"
+    "hdr-histogram-js": "^2.0.1",
+    "hdr-histogram-percentiles-obj": "^3.0.0",
+    "nice-napi": "^1.0.2"
```
eduardomourar: i believe this should be moved to the devDependencies

package.json
```diff
@@ -32,21 +32,22 @@
...
   },
   "optionalDependencies": {
```
eduardomourar, line 51: i have created PR #96 making `nice-napi` an optional peer dependency and it would be nice if those changes can be incorporated here

package.json
eduardomourar: or removed if using peer dependency approach

jasnell: good catch.. will get that updated

PR #96, opened 2020-11-18
Make nice-napi an optional peer dependency

eduardomourar:
In order to prevent `nice-napi` package from being installed, we will move it to peerDependency and then set the optional flag to true (only on NPM v6.11 or higher).


jasnell (approved):

jasnell:
Incorporated in #94

PR #103, opened 2020-12-14
Add MessagePort stream example

jasnell:


examples/stream/stream.mjs
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  _write (buf, _, cb) {
+    this[kPort].postMessage(buf, [buf.buffer]);
+    cb();
+  }
+
+  _writev (data, cb) {
+    for (let n = 0; n < data.length; n++) {
+      this[kPort].postMessage(data[n].chunk, [data[n].chunk.buffer]);
```
mcollina: I'd recommend to add batch support and pass the array through as-is.

examples/stream/stream.mjs
```diff
@@ -0,0 +1,103 @@
+import {
+  Writable,
+  Readable,
+  Duplex
+} from 'stream';
+
```
mcollina, line 6: All this files seems good utility that could live in a module.

examples/stream/stream.mjs
jasnell: Heh, yeah, I added that to the duplex, forgot to update the Writable.

examples/stream/stream.mjs
jasnell: Yeah, I'll eventually make it a separate module. Want to harden it up a bit more and work on some benchmarking first.

jasnell:
Btw, @mcollina @addaleax ... in the original version of this I had the worker using an async iterator to read the data off the inbound stream but for some reason it would not close/destroy properly. The _destroy() would call on the `MessagePort` to close but the close callback wouldn't fire. Using the streams API version doesn't have the same issue.

PR #122, opened 2021-04-30
Drop 10.x support, update deps, bump to 3.0.0

jasnell:
/cc @addalex @mcollina 

Updates deps and drops 10.x support. Bumps to 3.0.0.

~~@addaleax ... one of the updates here causes the `atomics-optimization.ts` test to timeout. If you have any ideas there let me know :-)~~

~~Also, the TAP update breaks coverage on some of the ESM related branches. I'm investigating. It's not clear if it's an issue with TAP or with our tests. ESM support does appear to work just fine.~~

Update: OK, so the coverage failures appear to be a limitation of TAP 15 to provide coverage on native ESM. Not surprising there but unfortunately that makes the coverage report a bit useless.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -8,10 +8,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Install Node.js 14
+      - name: Install Node.js 16
         uses: actions/setup-node@v1
         with:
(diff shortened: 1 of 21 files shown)


jasnell:
Ok, so, there's something in the `es2020` lib that causes the atomics optimization test to fail. Keeping lib at es2019 makes it work

mcollina (approved):
lgtm

PR #124, opened 2021-05-07
Add Piscina.prototype.run(), deprecate runTask()

jasnell:
The `runTask()` method uses positioned arguments for various
inputs, making the signature and the argument validation
fairly complicated. This makes it more difficult to add new
options later.

This commit introduces a new `run()` function that accepts an
options object instead, e.g.

```js
piscina.run({ a: 1 }, { filename: '...', signal: ... });
```

This will be more scalable in the long run if new options need
to be added.

This marks the `runTask()` variant as deprecated.

### DIFF
--- a/examples/abort/index.js
+++ b/examples/abort/index.js
@@ -11,7 +11,9 @@ const piscina = new Piscina({
 (async function () {
   const abortController = new AbortController();
   try {
-    const task = piscina.runTask({ a: 4, b: 6 }, abortController.signal);
+    const task = piscina.run(
+      { a: 4, b: 6 },
+      { signal: abortController.signal });
     abortController.abort();
     await task;
   } catch (err) {
--- a/examples/abort/index2.js
+++ b/examples/abort/index2.js
@@ -11,7 +11,7 @@ const piscina = new Piscina({
 (async function () {
   const ee = new EventEmitter();
   try {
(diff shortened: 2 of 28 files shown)


rubiin:
this looks good and completely readable considered to present imo, will it available on next release?

jasnell:
> ... will it available on next release?

That's the goal yes.

@addaleax and @mcollina ... this, btw, is partially in preparation for a new feature I'm working on that will allow multiple named worker functions to be exported from a single worker file.. e.g.

```js
// Worker

export async function foo() { /* ... */ }

export async function bar() { /* ... */ }
```

```js
piscina.run({ }, { name: 'foo' });
piscina.run({ }, { name: 'bar' });
```

rubiin:
> > ... will it available on next release?
> 
> That's the goal yes.
> 
> @addaleax and @mcollina ... this, btw, is partially in preparation for a new feature I'm working on that will allow multiple named worker functions to be exported from a single worker file.. e.g.
> 
> ```js
> // Worker
> 
> export async function foo() { /* ... */ }
> 
> export async function bar() { /* ... */ }
> ```
> 
> ```js
> piscina.run({ }, { name: 'foo' });
> piscina.run({ }, { name: 'bar' });
> ```

That would be so cool. maintaining different files has been a chore

addaleax (approved):

PR #128, opened 2021-05-15
Support multiple worker functions per file

jasnell:
Implements support for exporting multiple worker functions per worker.

worker.js:
```js
function add({ a, b }) { return a + b; }

function multiply({ a, b }) { return a * b; }

add.add = add;
add.multiply = multiply;

module.exports = add;
```

Then when running a task:

```js
piscina.run({ a: 2, b: 5 }, { name: 'add' });
piscina.run({ a: 2, b: 5 }, { name: 'multiply' });
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ const piscina = new Piscina({
 });
 
 (async function() {
-  const result = await piscina.runTask({ a: 4, b: 6 });
+  const result = await piscina.run({ a: 4, b: 6 });
   console.log(result);  // Prints 10
 })();
 ```
@@ -75,7 +75,7 @@ const piscina = new Piscina({
   filename: new URL('./worker.mjs', import.meta.url).href
 });
 
-const result = await piscina.runTask({ a: 4, b: 6 });
+const result = await piscina.run({ a: 4, b: 6 });
 console.log(result); // Prints 10
 ```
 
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -87,6 +87,43 @@ export default ({ a, b }) => {
...
+
+function multiply({ a, b }) { return a * b; }
+
+add.add = add;
+add.multiply = multiply;
+
+module.exports = { add, multiply };
+```
+
+The export to target can then be specifid when the task is submitted:
```
addaleax:
```suggestion
The export to target can then be specified when the task is submitted:
```

src/worker.ts
```diff
@@ -34,8 +34,8 @@ function getImportESM () {
 
 // Look up the handler function that we call when a task is posted.
 // This is either going to be "the" export from a file, or the default export.
-async function getHandler (filename : string) : Promise<Function | null> {
-  let handler = handlerCache.get(filename);
+async function getHandler (filename : string, name : string) : Promise<Function | null> {
+  let handler = handlerCache.get(`${filename}:${name}`);
```
addaleax:
```suggestion
  let handler = handlerCache.get(`${filename}/${name}`);
```

Because `:` is both a valid character in JS keys and POSIX filenames, so there’s at least some theoretical potential for collisions here?

src/worker.ts
```diff
@@ -65,7 +65,7 @@ async function getHandler (filename : string) : Promise<Function | null> {
     handlerCache.delete(key);
   }
 
-  handlerCache.set(filename, handler);
+  handlerCache.set(`${filename}:${name}`, handler);
```
addaleax:
```suggestion
  handlerCache.set(`${filename}/${name}`, handler);
```

README.md
```diff
@@ -87,6 +87,43 @@ export default ({ a, b }) => {
...
+
+```js
+'use strict';
+
+function add({ a, b }) { return a + b; }
+
+function multiply({ a, b }) { return a * b; }
+
+add.add = add;
+add.multiply = multiply;
```
addaleax, line 102: Are these two lines leftover from a previous implementation?

```suggestion
```

addaleax (dismissed):

rubiin:
This looks great. when can we expect this

README.md
```diff
@@ -87,6 +87,43 @@ export default ({ a, b }) => {
...
+
+module.exports = { add, multiply };
```
jasnell:
```suggestion
module.exports = add;
```

README.md
jasnell: No, actually the export line is wrong... It should export add as the default...

jasnell:
> This looks great. when can we expect this

I'm looking to do a release early next week.

addaleax (dismissed):

jasnell:
image

Oh... that's new and rather annoying.

addaleax (approved):
Well :)

PR #155, opened 2021-10-19
fix: transferrable not being transferred

Deivu:
I noticed that returning a `Transferable objects` from a worker doesn't get the object transferred, instead it's being cloned.

To verify this, I created my own Implementation of Transferrable
```js
class Shared {
    constructor({ ok, time, data } = {}) {
        this.ok = ok;
        this.time = time;
        this.data = data;
    }

    get [Piscina.transferableSymbol]() {
        return [this.data];
    }

    get [Piscina.valueSymbol]() {
        return { ok: this.ok, time: this.time, data: this.data };
    }

    make() {
        return Piscina.move(this);
    }
}

module.exports = Shared;
```
Then logged my buffer's byteLength intially, then after 1 second.

Without this fix, the byteLength didn't change, which signifies the buffer was not transferred
image
With the fix, the byteLength did change, which signifies the buffer was transferred
image

### DIFF
--- a/src/worker.ts
+++ b/src/worker.ts
@@ -146,7 +146,7 @@ function onMessage (
       }
       let result = await handler(task);
       if (isMovable(result)) {
-        transferList = result[kTransferable];
+        transferList = transferList.concat(result[kTransferable]);
         result = result[kValue];
       }
       response = {
--- a/src/worker.ts
+++ b/src/worker.ts
@@ -138,7 +138,7 @@ function onMessage (
 
   (async function () {
     let response : ResponseMessage;
-    const transferList : any[] = [];
+    let transferList : any[] = [];
     try {
       const handler = await getHandler(filename, name);
       if (handler === null) {
@@ -146,7 +146,7 @@ function onMessage (
       }
       let result = await handler(task);
       if (isMovable(result)) {
-        transferList.concat(result[kTransferable]);
+        transferList = result[kTransferable];
         result = result[kValue];
       }
       response = {
--- a/test/fixtures/send-buffer-then-get-length.js
+++ b/test/fixtures/send-buffer-then-get-length.js
@@ -1,0 +1,19 @@
+'use strict';
+
+const Piscina = require('../../dist/src');
+
+const response = {};
+module.exports = {
+  send: async () => {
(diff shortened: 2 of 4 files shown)


addaleax:
@Deivu Is this something that you could add a test for?

Deivu:
@addaleax  Took me some time to get the test working, but I did manage to add a test for it

arielszn:
Hey, we tried using the `Piscina.move()` feature for transferring a big object we need to deserialise on a worker thread, and then again a `Piscina.move()` from the worker thread to transfer back the deserialised object to the main thread, but this is not impacting on memory usage and latency as we were expecting for this change, we think it's not behaving as expected because of the issue that @Deivu is solving here.
Do you plan to merge this soon?

jasnell (approved):

jasnell:
Hmm looks like we've got a CI problem. I'll have to investigate that this weekend

sogame:
Hi @jasnell, I'm also interested in this fix for my project, any luck with the CI issues? 😬

arielszn:
> Hey, we tried using the `Piscina.move()` feature for transferring a big object we need to deserialise on a worker thread, and then again a `Piscina.move()` from the worker thread to transfer back the deserialised object to the main thread, but this is not impacting on memory usage and latency as we were expecting for this change, we think it's not behaving as expected because of the issue that @Deivu is solving here. Do you plan to merge this soon?

Hey, any news over here :) ?

nicholas-l:
Hey all, I have identified the issue with CI and have created a quick fix: https://github.com/piscinajs/piscina/pull/161

sogame:
Hi @Deivu, now that the CI issue has been solved (in https://github.com/piscinajs/piscina/pull/161), could you update your branch with `master`, so the new build is triggered? 😇

sogame:
Hi @jasnell would it be possible to get this fix merged (and even release a new version of `piscina` with it), please? 😇 
Really looking forward to test my project with this improvement 🚀

jasnell:
Running CI now, let's see where that lands first!

jasnell:
merged! I'll look at cutting a new release tomorrow afternoon

PR #161, opened 2021-11-28
Fix CI installing latest typescript

nicholas-l:
Currently the latest typescript is installed without a package-lock.json file. This file should be checked in so that the whole dependency tree is deterministic (see: https://nodejs.dev/learn/the-package-lock-json-file). Dont worry as this package-lock will not be used by npm when this package is used as a dependency in other packages.
This PR pins `typescript` to 4.3.x as typescript 4.4 has a semi-breaking change where they change `catch` argument from error to `unknown` (see: https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 .nyc_output
 .vscode
 node_modules
-package-lock.json
 dist
 coverage
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,13689 @@
+{
+  "name": "piscina",
+  "version": "3.1.0",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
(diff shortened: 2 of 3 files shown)


nicholas-l:
I have also created another PR so that dependency updates to package.json and package-lock.json are automated by dependabot: https://github.com/piscinajs/piscina/pull/162 now that the dependencies are locked.

jasnell (approved):

nicholas-l:
I think `Test (ubuntu-latest, 13.x)` check is never going to complete as it isn't in the test matrix, and it is marked as required. I believe you can remove this required test in the branch protection rules.

PR #163, opened 2021-12-06
feat(atomics): add environment variable to disable atomics

SamVerschueren:
Hi 👋 !

As discussed privately, it would be nice for runtimes to disable the use of `Atomics` globally as it does not change the outcome at all. A runtime like WebContainer runs in the browser and is currently not capable of implementing `receiveMessageOnPort` as the browser does not have such implementation.

By adding this environment variable, we are capable of turning it off in general and make piscina work out-of-the-box in StackBlitz WebContainers.

Feel free to suggest different wording for the readme.

Sam

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,7 +303,8 @@ This class extends [`EventEmitter`][] from Node.js.
     to the task. Keep in mind that Worker threads are generally not built for
     handling I/O in parallel.
   * `useAtomics`: (`boolean`) Use the [`Atomics`][] API for faster communication
-    between threads. This is on by default.
+    between threads. This is on by default. You can disable `Atomics` globally by
+    setting the environment variable `PISCINA_DISABLE_ATOMICS` to `1`.
   * `resourceLimits`: (`object`) See [Node.js new Worker options][]
     * `maxOldGenerationSizeMb`: (`number`) The maximum size of each worker threads
       main heap in MB.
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -1,7958 +1,8 @@
 {
   "name": "piscina",
   "version": "3.1.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
jasnell: Please pull the lock file change back out of this PR :-)

jasnell (dismissed):

package-lock.json
SamVerschueren: Woops. In all the excitement, I totally looked over these changes 😂 . Thanks!

SamVerschueren:
I've also replaced `eval()` with `new Function()` as that's the preferred way of evaluating as explained in the [Never use eval!](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval#never_use_eval!) section.

d3lm (dismissed):

jasnell (dismissed):

SamVerschueren:
Sorry. I just saw that I had to disable a lint rule for the `new Function()`. Fixed it and verified that linting is now correct.

d3lm (approved):

jasnell (approved):

PR #169, opened 2021-12-11
Update dependencies

jasnell:


jasnell:
@addaleax ... any ideas why the dependency update here would break the atomics test?

PR #213, opened 2022-03-27
Add communication from worker to main thread

### LINKED ISSUE #212: Workers to main thread communication
I need to add a way to send messages from workers to main thread:

The only possibility I see right now with Piscina, is to send a messagePort when running a task and use it to communicate (Or use a broadcastChannel but only for node 17+).

In my case, I want to use a single messagePort per worker, set at startup of the worker and use it when initializing inside the worker and before running tasks... But there is no currently easy way to add a messagePort per worker at startup built in Piscina...

I wonder if it would be possible to use the already existing built in messages ports of Piscina to add this feature ?

Something like this:

Adding a new kind of message that the worker can send:
```js
export interface GenericMessage {
  data: any;
}
```

Use it to send messages from worker at any stage:
```js
let message : GenericMessage;
parentPort.postMessage(message);
```

And handle such messages at pool level to expose it:
```js
worker.on('message', (message : GenericMessage) => {
  publicInterface.emit('message', message.data);
}
```

Do this make sense ?

Elyahou:
Fix: https://github.com/piscinajs/piscina/issues/212

### DIFF
--- a/README.md
+++ b/README.md
@@ -418,6 +418,10 @@ itself.
 
 A `'drain'` event is emitted whenever the `queueSize` reaches `0`.
 
+### Event: `'message'`
+
+A `'message'` event is emitted whenever a message is received from a worker thread.
+
 ### Property: `completed` (readonly)
 
 The current number of completed tasks.
--- a/examples/messages/index.js
+++ b/examples/messages/index.js
@@ -1,0 +1,16 @@
+'use strict';
+
+const Piscina = require('../../dist/src');
+const { resolve } = require('path');
+
+const piscina = new Piscina({
+  filename: resolve(__dirname, 'worker.js')
+});
+
+(async function () {
+  piscina.on('message', (event) => {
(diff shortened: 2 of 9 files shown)


jasnell (approved):
LGTM but I'd like to get @addaleax's opinion also

Elyahou:
@jasnell @addaleax Any news ?

alolis:
@Elyahou, can you please provide instructions on how did you add "current" branch of piscina to your project in order to use your merged PR?

I am mostly interested with the steps AFTER the `npm/yarn add https://github.com/piscinajs/piscina.git#current` part ;)

metcoder95:
You can take a look at npm official docs [here](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#github-urls).
Recommend you stay tuned for possible upgrades/merges against the main branch as is possible that unexpected changes can break your app between installations 🙂

PR #239, opened 2022-06-09
fix: Fix types for TypeScript 4.7

BlackGlory:
The module's type definition file could not be found in some cases of TypeScript 4.7, this PR resolves the issue.

Reference: https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#package-json-exports-imports-and-self-referencing

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "A fast, efficient Node.js Worker Thread Pool implementation",
   "main": "./dist/src/index.js",
   "exports": {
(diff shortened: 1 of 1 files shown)


jasnell (approved):

BlackGlory:
Any progress?

drwpow:
This would fix a bug for me as well. I believe this is safe to merge without any downside (in general, it never hurts to expose more to `exports`, e.g. `"*": "./*"`).

PR #318, opened 2022-12-29
Option to create `WORKER_ID` env var in threads

overlookmotel:
This PR adds an option `workerIdEnv` which results in each thread having an environment variable set with the ID of the thread. NB The thread, not the task.

* `workerIdEnv: true` sets a var `process.env.WORKER_ID`
* `workerIdEnv: 'FOO_ID'` sets a var `process.env.FOO_ID`
* Default behavior is to do nothing (no breaking change)

Use case: Debugging running different tasks across threads. If one task is poisoning the thread environment somehow and causing failures later in unrelated tasks run in the same thread, it's helpful to understand what is running in which thread.

I've run into exactly this problem while using [jest-light-runner](https://www.npmjs.com/package/jest-light-runner) which runs each test file in a Piscina pool. One test was passing, but it left `global` in a bad state which caused failures in other unrelated test files, purely by the accident that they ran on same thread. Figuring out the source of the fault was difficult without being able to narrow down which other test files ran on the same thread as the failing test.

Two queries about the content of this PR:

1. I'm not a TypeScripter and had difficulty with the type checker on one line. Perhaps someone can advise how to get around this? https://github.com/overlookmotel/piscina/blob/worker-ids/src/index.ts#L586-L591
2. Query: If `options.env === undefined`, this currently copies the main thread's `process.env` before adding `WORKER_ID`. Inheriting parent's env is the default behavior with worker threads, right?

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,6 +338,9 @@ This class extends [`EventEmitter`][] from Node.js.
     `fs.close()`, and will close them automatically when the Worker exits.
     Defaults to `true`. (This option is only supported on Node.js 12.19+ and
     all Node.js versions higher than 14.6.0).
+  * `workerIdEnv`: (`boolean|string`) If `true`, sets an environment variable in
+    worker threads `WORKER_ID` with the ID of each thread (`process.env.WORKER_ID`).
+    If a string, that name is used instead of `WORKER_ID`.
 
 Use caution when setting resource limits. Setting limits that are too low may
 result in the `Piscina` worker threads being unusable.
--- a/src/common.ts
+++ b/src/common.ts
@@ -2,6 +2,8 @@ import type { MessagePort } from 'worker_threads';
 
 export const READY = '_WORKER_READY';
 
+export const DEFAULT_WORKER_ID_ENV = 'WORKER_ID';
+
 export interface StartupMessage {
   filename : string | null;
   name : string;
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,6 +12,7 @@ import { performance } from 'perf_hooks';
 import hdrobj from 'hdr-histogram-percentiles-obj';
 import {
   READY,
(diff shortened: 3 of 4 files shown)


metcoder95:
I've not been able to fully unblock most of the PRs 😞

README.md
```diff
@@ -360,6 +360,9 @@ This class extends [`EventEmitter`][] from Node.js.
     `fs.close()`, and will close them automatically when the Worker exits.
     Defaults to `true`. (This option is only supported on Node.js 12.19+ and
     all Node.js versions higher than 14.6.0).
+  * `workerIdEnv`: (`boolean|string`) If `true`, sets an environment variable in
```
metcoder95, line 359: I think that we should start simple, just using `workerId` should be enough to create the environment var with the id of the worker within.

Customizing the name might be interesting, but I'll rather prefer a separated API for it, maybe `workerIdName` or similar.

metcoder95 (commented):
> I'm not a TypeScripter and had difficulty with the type checker on one line. Perhaps someone can advise how to get around this? [overlookmotel/piscina@worker-ids/src/index.ts#L586-L591](https://github.com/overlookmotel/piscina/blob/worker-ids/src/index.ts?rgh-link-date=2022-12-29T21%3A25%3A31Z#L586-L591)

- What's the error you're facing exactly?

> Query: If options.env === undefined, this currently copies the main thread's process.env before adding WORKER_ID. Inheriting the parent's env is the default behavior with worker threads, right?

Yes 👍 

Feature LGTM, the only concern is to standardize the value of the `WORKER_ID`. IMO we should stick to pure strings, it does not matter that is an auto incremental number, it is always passed as a string.

metcoder95:
Closing due to inactivity, feel free to open a fresh one if interested in continuing the work 👍

PR #342, opened 2023-03-31
docs: document out of scope behavior

### LINKED ISSUE #316: Unawaited promises behavior
Hi,

I am seeing a strange behavior that may be coming from a misunderstanding of how piscina or worker threads behave.

In a scenario where the worker triggers some async task that is not awaited, it seems that this task is "suspended", and will be resumed the next time the worker wakes up.

I have put up a reproduction [here](https://gist.github.com/Marsup/998cffff259fff7ae87360bb9afb4135). Notice how the "Resolved" logs appear at the beginning of the next run of the worker and not 500ms after it's finished. This is a trivial case with timers, but it happens for us with http (axios) calls.

What am I missing here?

Marsup:
Following the discussion in #316, I think it's a good clarification to add to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,6 +252,25 @@ stream
   });
 ```
 
+### Out of scope asynchronous code
+
+A worker thread is **only** active from the moment you call it, to the moment it returns a result. Once this is done, Piscina will wait for stdout and stderr to be flushed, and then pause the worker's event-loop until the next call. If async code is scheduled without being awaited before returning, since Piscina has no way of detecting this, that code execution will be resumed on the next call. Thus, it is highly recommended to properly handle all async tasks before returning a result as it could make your code unpredictable.
+
+For example:
+
+```js
+const { setTimeout } = require('timers/promises');
+
+module.exports = ({ a, b }) => {
+  // This promise should be awaited
+  setTimeout(1000).then(() => {
+    console.log('Working'); // This will **not** run during the same worker call
+  });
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -252,6 +252,25 @@ stream
   });
 ```
 
+### Out of scope asynchronous code
+
+A worker thread is **only** active from the moment you call it, to the moment it returns a result. Once this is done, Piscina will wait for stdout and stderr to be flushed, and then pause the worker's event-loop until the next call. If async code is scheduled without being awaited before returning, since Piscina has no way of detecting this, that code execution will be resumed on the next call. Thus, it is highly recommended to properly handle all async tasks before returning a result as it could make your code unpredictable.
```
metcoder95:
```suggestion
A worker thread is **only** active until the moment it returns a result, it can be a result of a synchronous call or a Promise that will be fulfilled/rejected in the future. Once this is done, Piscina will wait for stdout and stderr to be flushed, and then pause the worker's event-loop until the next call. If async code is scheduled without being awaited before returning since Piscina has no way of detecting this, that code execution will be resumed on the next call. Thus, it is highly recommended to properly handle all async tasks before returning a result as it could make your code unpredictable.
```

metcoder95 (approved):

metcoder95:
I've not been able to fully unblock most of the PRs 😞

PR #344, opened 2023-04-13
correctly handle process.exit calls outside of a task

### LINKED ISSUE #304: Doesn't catch process.exit() in the worker thread
If a worker thread calls process.exit(), it doesn't seem like Piscina catches that, i.e. it doesn't restart the worker thread or anyhow marks it as "non-existing". I.e. the next job scheduled may be sent to that "zombie" worker and stuck there forever unnoticed.

I found out that this code: https://github.com/piscinajs/piscina/blob/current/src/index.ts#L668 triggers in such a situation, but it doesn't remove the thread from the pool nor does anything else.

Notice that if the worker thread is terminated on an error (e.g. with `throw Error("ouch")` in its setTimeout() handler, i.e. by producing a fatal uncaughtException), this is handled correctly. The only issue is with a worker exiting "successfully" (and this may actually happen due to some inaccurate written code or something).

Also, worker_threads supports on("exit") event, so there IS a way to catch process.exit() from the worker thread.

clayjones-at:
While using piscina, we discovered that `process.exit()` calls within workers during the execution of tasks are not yet explicitly handled. When a task's execution results in `process.exit()`, the task and any others assigned to the worker thread do not get marked done. Also, a new worker is not added to the pool, which may lead to the pool drying up over time. This PR is an attempt to fix this behavior by erroring out any remaining tasks assigned to a worker when it exits, and restarting the worker. 

Note also that behavior is a little different when `useAtomics: false`, for reasons I didn't take the time to fully understand.

Attempts to fix https://github.com/piscinajs/piscina/issues/304

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -635,33 +635,14 @@ class ThreadPool {
     });
 
     worker.on('error', (err : Error) => {
-      // Work around the bug in https://github.com/nodejs/node/pull/33394
-      worker.ref = () => {};
+      this._onError(worker, workerInfo, err, false);
+    });
 
-      // In case of an uncaught exception: Call the callback that was passed to
-      // `postTask` with the error, or emit an 'error' event if there is none.
-      const taskInfos = [...workerInfo.taskInfos.values()];
-      workerInfo.taskInfos.clear();
-
-      // Remove the worker from the list and potentially start a new Worker to
-      // replace the current one.
-      this._removeWorker(workerInfo);
-
-      if (workerInfo.isReady() && !this.workerFailsDuringBootstrap) {
-        this._ensureMinimumWorkers();
-      } else {
-        // Do not start new workers over and over if they already fail during
-        // bootstrap, there's no point.
-        this.workerFailsDuringBootstrap = true;
-      }
-
-      if (taskInfos.length > 0) {
-        for (const taskInfo of taskInfos) {
-          taskInfo.done(err, null);
-        }
-      } else {
-        this.publicInterface.emit('error', err);
-      }
+    worker.on('exit', (exitCode : number) => {
+      const err = new Error(`worker exited with code: ${exitCode}`);
+      // Only error unfinished tasks on process exit, since there are legitimate
+      // reasons to exit workers and we want to handle that gracefully when possible.
+      this._onError(worker, workerInfo, err, true);
     });
 
     worker.unref();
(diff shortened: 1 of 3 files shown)


clayjones-at:
I'm about to be afk for a couple weeks, so I can't immediately respond to comments on this PR, though I do intend to see this change through if maintainers actually think it's a good idea. If others want to take this PR from me in the meantime (or in general) I'd love that!

src/index.ts
```diff
@@ -635,33 +635,14 @@ class ThreadPool {
...
 
-      if (taskInfos.length > 0) {
-        for (const taskInfo of taskInfos) {
-          taskInfo.done(err, null);
-        }
-      } else {
-        this.publicInterface.emit('error', err);
-      }
+    worker.on('exit', (exitCode : number) => {
+      const err = new Error(`worker exited with code: ${exitCode}`);
```
metcoder95, line 642: Maybe here we can distinguish based on the exit code.
If the user intends to gracefully shutdown its thread, it can do it by sending `0` as the exit code, otherwise when due to a failure, `1` will be returned, then it can be wrapped inside of an Error object, and Piscina can bootup a new thread.

It will also require to adjust documentation to highlight that

clayjones-at: Hmm, yeah I hadn't thought about a `0` exit code. That said, this code only errors out any existing unfinished tasks, as opposed to emitting an error over the `publicInterface` itself.

I'm not sure piscina should accept a `0` exit with unfinished tasks more gracefully than it already does in this PR - but maybe we'd rather silently retry the tasks once on a new thread, or something similar? I'm not familiar enough with piscina to have an opinion here; deferring to you.

metcoder95: Hmm, no I think you're right, handling all of the errors and wrapping them for allowing piscina create a new thread sounds good 👍

metcoder95 (approved):

clayjones-at:
Is there anything else I need to do to merge, or is that up to you/other maintainers?

metcoder95:
There's nothing else to add, it seems there are some issues with the CI that are blocking the merge of some PRs. We'll merge it as soon as we unblock the issue 🙂

clayjones-at:
sounds good, thanks!

metcoder95:
@clayjones-at can you verify, it seems is failing over v16 🤔

clayjones-at:
On my mac (OS 13.2.1 (22D68), M1), the following commands work. Does this verify? 
```
sudo n 18.16.0
alias node=/usr/local/bin/node
alias npm=/usr/local/bin/npm
npm run tests
```
I'm a little suspicious of the OS/CI setup given just the mac runs are failing. Pretty new to node still, so not sure what the common cross-OS gotchas are.

metcoder95:
I've not being able to fully unblock most of the PRs 😞

clayjones-at:
Not a problem, [done](https://github.com/piscinajs/piscina/pull/361)

PR #349, opened 2023-06-09
chore: Update CI

metcoder95:
The PR aims to update the CI manifest and `dependabot` scope and cadency to up-to-date settings.
It introduces the following changes:

- chore: update CI
- chore: update dependabot manifest
- chore: sync lockfile

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -25,7 +25,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         node-version: [12.x, 14.x, 16.x]
-    runs-on: ${{matrix.os}}
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
         with:
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -24,7 +24,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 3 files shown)


metcoder95:
Reverted to `12`, `14`, and `16` as it seems there are changes required to support `18` onwards. I'll address it in a subsequent PR targeting a new branch for dropping support for old Node versions.

cc: @jasnell @RafaelGSS

.github/workflows/nodejs.yml
```diff
@@ -8,30 +8,43 @@ jobs:
...
   test:
+    needs: [lint]
     name: Test
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node-version: [12.x, 14.x, 16.x, 17.x]
-    runs-on: ${{matrix.os}}
+        node-version: [12.x, 14.x, 16.x]
```
RafaelGSS: Why don't drop 12 and 14 here?

metcoder95: I was considering it to add the whole drop of v12-14 within the work on `next` branch, as I think the drop should also sync with a drop of the supported version within `engines` of `package.json`. Wdyt?

metcoder95:
@RafaelGSS, do you have an idea of from where the `13` is coming?
I tried to see where this is coming from but couldn't reach a conclusion. The `nodejs.yml` does not contains the v13 within the matrix

RafaelGSS (commented):
@metcoder95 I think you can target next and drop v12, v14. So I can release a new major too

metcoder95:
SGTM, it's done now 👍

PR #361, opened 2023-06-15
feat: correctly handle process.exit calls outside of a task

### LINKED ISSUE #304: Doesn't catch process.exit() in the worker thread

clayjones-at:
clone of [this PR](https://github.com/piscinajs/piscina/pull/344) made to help avoid some CI troubles

Attempts to fix #304

### DIFF
--- a/test/test-is-buffer-transferred.ts
+++ b/test/test-is-buffer-transferred.ts
@@ -9,7 +9,7 @@ function wait () {
 test('transferable objects must be transferred', async ({ equal }) => {
   const pool = new Piscina({
     filename: resolve(__dirname, 'fixtures/send-buffer-then-get-length.js'),
-    useAtomics: false
+    // useAtomics: false // let's see if turning on atomics makes a difference
   });
   await pool.run({}, { name: 'send' });
   await wait();
--- a/test/test-is-buffer-transferred.ts
+++ b/test/test-is-buffer-transferred.ts
@@ -8,7 +8,7 @@ function wait () {
 
 test('transferable objects must be transferred', async ({ equal }) => {
   const pool = new Piscina({
-    filename: resolve(__dirname, 'fixtures/send-buffer-then-get-length.js'),
+    filename: resolve(__dirname, 'fixtures/send-buffer-then-get-length.js')
     // useAtomics: false // let's see if turning on atomics makes a difference
   });
   await pool.run({}, { name: 'send' });
--- a/test/test-is-buffer-transferred.ts
+++ b/test/test-is-buffer-transferred.ts
@@ -8,8 +8,8 @@ function wait () {
 
 test('transferable objects must be transferred', async ({ equal }) => {
   const pool = new Piscina({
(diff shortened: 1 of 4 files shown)


metcoder95:
The issue is quite interesting, I've tried to isolate it and see what can be done but it works locally at the three different versions it fails but it succeeds on all of them.

I'll give it a try with [`act`](https://t.co/5KbhL0xvTg) to see if I can find a way to isolate them, recommend you give it a try as well 🙂

clayjones-at:
Ok, yeah I will look more seriously into this hopefully at some point this week

clayjones-at:
Pardon the incoming commits, while I fiddle with getting `act` working I'm trying some things in CI. Not planning on rapid-firing my commits so hopefully that won't cause much/any cost increase

test/test-is-buffer-transferred.ts
```diff
@@ -8,8 +8,8 @@ function wait () {
 
 test('transferable objects must be transferred', async ({ equal }) => {
   const pool = new Piscina({
-    filename: resolve(__dirname, 'fixtures/send-buffer-then-get-length.js'),
-    useAtomics: false
+    filename: resolve(__dirname, 'fixtures/send-buffer-then-get-length.js')
+    // useAtomics: false // let's see if turning on atomics makes a difference
```
metcoder95: This is the culprit? 🤔

clayjones-at: oh _wow_, I stopped paying attention to this over the weekend and kept getting failures in act, so didn't notice tests pass.

There's no way this specific line change was the culprit, since more than just this test failed last run. Since tests pass anyways I have no hypothesis anymore. Have you observed your CI tests flake like this in the past?

metcoder95: Not really, we had some issues when migrating to v20, but I don't think that should be the culprit then 🤔 

Can you uncomment it and see if that solves the issue?

clayjones-at: sure, done. let's see!

clayjones-at: well, i'll take it!

metcoder95 (approved):
Nice! We can finally merge it! Thanks for the patience 🙇

PR #363, opened 2023-06-19
refactor!: drop runTask method

metcoder95:
## Context
The PR attempts to drop the `runTask` method signature as part of the new `v5` work.

## Changes
- refactor: replace abortcontroller lib with native
- refactor!: remove deprecated runTask method

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,33 +389,6 @@ an error, the returned `Promise` will be rejected with that error.
 If the task is aborted, the returned `Promise` is rejected with an error
 as well.
 
-### Method: `runTask(task[, transferList][, filename][, abortSignal])`
-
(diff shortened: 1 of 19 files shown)


RafaelGSS (commented):
Can you elaborate on why should we do it? I might be missing something due to context switching.

metcoder95:
`Piscina#runTask` has been marked as deprecated for a while. According to the commit history, it seems that we wanted to reduce the number of positional arguments and opted for an object passed as a second argument instead. See: https://github.com/piscinajs/piscina/commit/a58f99a350574f4c40fa6f322a9d229587b53f79

I'm pointing to `next` and seeking to remove the whole method

RafaelGSS (approved):
LGTM!

Issue #126, opened 2021-05-12
Add `needDrain` property


ronag:
The current way to determine whether a pool needs a drain event before queueing more tasks is a little awkward. Can we add a `needDrain` proprerty?

```js
if (pool.needDrain) {
  await EE.once(pool, 'drain')
}
pool.runTask()
```

ronag:
When does `'drain'` actually get emitted? Is it when the queue is empty or when the queue.size < queue.capacity?

jasnell:
Currently when the queue is empty. The plan has been to go back and revisit that but haven't been able to get to it yet.

ronag:
That’s doesn’t sound optimal.

jasnell:
Yeah, it's not the most optimal approach but I haven't really had the opportunity to go back and figure out a more optimal strategy.

metcoder95:
Which should be the optimal approach?

metcoder95:
Hey @ronag, I'm planning to revisit this, do you have any suggestion for an optimal approach?

ronag:
Not really. But I do think https://github.com/piscinajs/piscina/issues/348 is a duplicate of this.

metcoder95:
I believe that we can have two action paths here:

1.  We can maybe start by adjusting the `drain` method to be triggered once as soon as the `queue.size < queue.capacity`, in that way we enable continuing work as soon as there is spare capacity to schedule more tasks (this will translate into a major, so we will need to point to `next` branch for this one)
2. Adding the `needsDrain`can be used as soon as the `maxQueue` has been reached. That should cover the requirement to understand when the pool needs to be waited for being drained until schedule a new task

PR #367, opened 2023-06-30
refactor: emit `drain` as soon as capacity is available

metcoder95:
Implements insights from #126.

The new rationale to emit `drain` event, goes as follows:

>The rationale is to checkpoint the capacity every time new
>task is being schedulded or appended to the queue.
>If capacity exceeded but we having checkpointed yet the check,
>we allow the pool to see if it can flush itself soon.
>
>At subsequent checks, we verify that the previously checkpoint
>was negative, and its current capacity. If the capacity excheded
>and previous checkpoint was negative, and the capacity is above
>the workload, we emit the drain event.
>
>If previous checkpoint is false, but the capacity stills exceeded,
>we checkpoint as false, and wait for a further checkpoint to validate
>once more the checks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,7 +434,8 @@ itself.
 
 ### Event: `'drain'`
 
-A `'drain'` event is emitted whenever the `queueSize` reaches `0`.
+A `'drain'` event is emitted whenever the `queueSize` is below the
+total capacity of the pool.
 
 ### Event: `'needsDrain'`
 
--- a/README.md
+++ b/README.md
@@ -434,8 +434,11 @@ itself.
 
 ### Event: `'drain'`
 
-A `'drain'` event is emitted whenever the `queueSize` is below the
-total capacity of the pool.
+A `'drain'` event is emitted when the current usage of the
+pool is below the maximum capacity of the same.
+The intended goal is to provide backpressure to the task source
+so creating tasks that can not be executed at immediately can be avoided.
+
 
 ### Event: `'needsDrain'`
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -526,6 +526,7 @@ class ThreadPool {
   inProcessPendingMessages : boolean = false;
   startingUp : boolean = false;
   workerFailsDuringBootstrap : boolean = false;
+  checkpointStatus: boolean = true;
 
   constructor (publicInterface : Piscina, options : Options) {
     this.publicInterface = publicInterface;
@@ -883,9 +884,30 @@ class ThreadPool {
     const totalCapacity = this.options.maxQueue + this.pendingCapacity();
     const totalQueueSize = this.taskQueue.size + this.skipQueue.length;
 
-    if (totalQueueSize === 0) {
-      this.needsDrain = false;
+    /**
+     * The rationale is to checkpoint the capacity every time new
+     * task is being schedulded or appended to the queue.
+     * If capacity exceeded but we having checkpointed yet the check,
+     * we allow the pool to see if it can flush itself soon.
+     *
+     * At subsequent checks, we verify that the previously checkpoint
+     * was negative, and its current capacity. If the capacity excheded
+     * and previous checkpoint was negative, and the capacity is above
+     * the workload, we emit the drain event.
+     *
+     * If previous checkpoint is false, but the capacity stills exceeded,
+     * we checkpoint as false, and wait for a further checkpoint to validate
+     * once more the checks.
+     */
+    const isCapacityExceeded = totalQueueSize >= totalCapacity;
+    const shouldEmitDrain = !isCapacityExceeded && !this.checkpointStatus;
+
+    if (shouldEmitDrain) {
       this.publicInterface.emit('drain');
+    } else if (isCapacityExceeded) {
+      this.checkpointStatus = false;
+    } else {
+      this.checkpointStatus = true;
     }
 
     if (totalQueueSize >= totalCapacity) {
--- a/src/index.ts
+++ b/src/index.ts
@@ -532,7 +532,7 @@ class ThreadPool {
   completed : number = 0;
   runTime : Histogram;
   waitTime : Histogram;
-  needsDrain : boolean;
+  _needsDrain : boolean;
   start : number = performance.now();
   inProcessPendingMessages : boolean = false;
   startingUp : boolean = false;
@@ -571,7 +571,7 @@ class ThreadPool {
     this.startingUp = true;
     this._ensureMinimumWorkers();
     this.startingUp = false;
-    this.needsDrain = false;
+    this._needsDrain = false;
   }
 
   _ensureMinimumWorkers () : void {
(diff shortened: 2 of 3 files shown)


RafaelGSS:
is it different from #368?

metcoder95:
>is it different from https://github.com/piscinajs/piscina/pull/368?

Yes, as this change affects the way we decide when to emit the `drain` event. #368 is meant to add a new property only.

I'll rebase the other one on top of this one once merged

README.md
```diff
@@ -438,7 +438,8 @@ itself.
 
 ### Event: `'drain'`
 
-A `'drain'` event is emitted whenever the `queueSize` reaches `0`.
+A `'drain'` event is emitted whenever the `queueSize` is below the
```
RafaelGSS: I'm not sure we should do it. The docs clearly show an example:

> The 'drain' event may be used to receive notification when the queue is empty and all tasks have been submitted to workers for processing. -- https://github.com/piscinajs/piscina#backpressure

metcoder95: That's a good point, and the current approach might not be the more optimal as it requires waiting until the queue is totally empty before resuming work, which can not be the best depending on the workload (e.g. extending waiting times, etc.)

The idea was to allow the pool to notify whenever there is spare capacity to be used for scheduling new tasks, maybe we can think of something better?

ronag: IMHo this should work the same as `stream.Writable` or be called something else. Otherwise it is confusing. What `'drain'` means is kind of established.

RafaelGSS: If we change this behaviour, it should be a semver-major.

@ronag I couldn't find it in the docs, but the behaviour stream.Writable does it the same as this PR is trying to apply?

metcoder95: >If we change this behaviour, it should be a semver-major.

Sure, is already marked as so and pointing to `next` branch 👍 

>@ronag I couldn't find it in the docs, but the behaviour stream.Writable does it the same as this PR is trying to apply?

Yeah, have the same issue when trying to check `stream.Writable`, so how we should move forward? 🙂

RafaelGSS: @mcollina @jasnell any thoughts on this?

ronag:
Giving this some more thought here is what I currently think is the "correct" behaviour (inspired by looking at Writable).

Define `pending` as number of tasks in-flight.
Define `maxPending ` as `maxThreads * concurrentTasksPerThread`

Our goal is to make it possible for user space to use the pool in a way where always `waiting === 0`, since we want to avoid creating tasks that can't execute immediately in order to provide back pressure to the task source.

Hence, our logic should be:

Whenever, a task is added AND `pending === maxPending`, set `needsDrain` to `true`.
Whenever, a task finishes executing AND `pending < maxPending `, set `needsDrain` to `false` and emit `'drain'`.

EDIT: Fixed incorrect logic.

metcoder95:
>Our goal is to make it possible for user space to use the pool in a way where always waiting === 0, since we want to avoid creating tasks that can't execute immediately in order to provide back pressure to the task source

Got it, this was the part I was missing

>Whenever, a task is added AND pending === maxPending, set needsDrain to true.
>Whenever, a task finishes executing AND waiting === 0, set needsDrain to false and emit 'drain'.

That resonates to me, then for #368, I'll change the order of the `needDrain` set, and rebase on top of this branch to implement what we discussed about

ronag:
Actually, change this:

> Whenever, a task finishes executing AND waiting === 0, set needsDrain to false and emit 'drain'.

To

> Whenever, a task finishes executing AND  pending < maxPending set needsDrain to false and emit 'drain'.

metcoder95:
Applied the change, it has a dependency on #368, so I'll add the remaining bits after that one is merged 👍

src/index.ts
```diff
@@ -874,7 +885,17 @@ class ThreadPool {
...
+     * Our goal is to make it possible for user space to use the pool
+     * in a way where always waiting === 0, since we want to avoid creating tasks that can't execute
+     * immediately in order to provide back pressure to the task source.
+     */
+    const maxCapacity = this.options.maxThreads * this.options.concurrentTasksPerWorker;
+    const currentUsage = this.workers.getCurrentUsage();
+
+    if (maxCapacity === currentUsage) {
+      // TODO: needs drain goes here
+    } else if (maxCapacity > currentUsage) {
```
ronag, line 898:
```suggestion
    } else if (maxCapacity > currentUsage && needsDrain) {
```

ronag, line 898:
```suggestion
    /**
     * Our goal is to make it possible for user space to use the pool
     * in a way where always waiting === 0, since we want to avoid creating tasks that can't execute
     * immediately in order to provide back pressure to the task source.
     */
    const maxCapacity = this.options.maxThreads * this.options.concurrentTasksPerWorker;
    const currentUsage = this.workers.getCurrentUsage();

    if (currentUsage  >= maxCapacity && !this._needsDrain) {
      this._needsDrain = true
    } else if (currentUsage < maxCapacity && this._needsDrain) {
```

```suggestion
    if (maxCapacity >= currentUsage) {
      this._needsDrain = true
    } else if (maxCapacity > currentUsage && this._needsDrain) {
```

src/index.ts
```diff
@@ -374,6 +374,17 @@ class AsynchronouslyCreatedResourcePool<
   onAvailable (fn : (item : T) => void) {
     this.onAvailableListeners.push(fn);
   }
+
+  getCurrentUsage (): number {
+    let inFlight = 0;
+    for (const worker of this.readyItems) {
+      const currentUsage = worker.currentUsage();
+
+      if (currentUsage < Infinity) inFlight += currentUsage;
```
```suggestion
      if (Number.isFinite(currentUsage)) inFlight += currentUsage;
```

README.md
ronag: This description needs to be updated with the current intended behaviour.

src/index.ts
```diff
@@ -874,7 +885,17 @@ class ThreadPool {
   }
 
   _maybeDrain () {
-    if (this.taskQueue.size === 0 && this.skipQueue.length === 0) {
+    /**
+     * Our goal is to make it possible for user space to use the pool
+     * in a way where always waiting === 0, since we want to avoid creating tasks that can't execute
+     * immediately in order to provide back pressure to the task source.
+     */
+    const maxCapacity = this.options.maxThreads * this.options.concurrentTasksPerWorker;
```
ronag: I don't think the options object is immutable so it might be a good idea to recalculate and store this value.

metcoder95: Maybe I'll be missing something, but this is not exposed through the `Piscina` interface, in fact, the `ThreadPool` is hidden under a private variable, so it should be safe enough and in theory not mutable (at least not at user-land).

Or do you refer to store the `maxCapacity` to avoid calculate it every time?

src/index.ts
metcoder95: Waiting for #368 👍

src/index.ts
ronag: I don't actually know where options comes from so maybe it's not a problem...

metcoder95: Where the `pool` is being stored: https://github.com/piscinajs/piscina/blob/0af59c6b866a25cfda45fbd849a98e8ca3982606/src/index.ts#L925-L926

Where `options` are passed: https://github.com/piscinajs/piscina/blob/0af59c6b866a25cfda45fbd849a98e8ca3982606/src/index.ts#L984

It should be safe, I believe

ronag: Actually... no... the `options` object comes from the `Piscina` constructor which is a user provided object and can be modified externally by the user.

ronag: That's probably a general problem. You should make sure there is no outside reference to internal state.

metcoder95: Ah, got it. Yes, you are right, as preserve the object down the path. We'll need to either store it somewhere or make a deep copy of it to avoid external mutations.

Most likely it will deserve its own PR.
For now, I'll store the value within the `ThreadPool` to avoid side-effects.

src/index.ts
metcoder95: Done in [f91a697](https://github.com/piscinajs/piscina/pull/367/commits/f91a6979c9f4342e15027ffe6ed9e67813fa90d6)

ronag (approved):

RafaelGSS (approved):

PR #368, opened 2023-06-30
feat: add `needsDrain` property

### LINKED ISSUE #126: Add `needDrain` property

metcoder95:
Closes #126

**Pendings**

- [x] Testing
- [x] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -503,6 +503,14 @@ An Array of the `Worker` instances used by this pool.
 
 The current number of tasks waiting to be assigned to a Worker thread.
 
+### Property: `needsDrain` (readonly)
+
+Boolean value that specifices wether the capacity of the pool has
+been exceeded by the number of tasks submitted.
+
+This property is helpful to make decisions towards creating backpressure
+over the number of tasks submitted to the pool.
+
 ### Property: `utilization` (readonly)
 
 A point-in-time ratio comparing the approximate total mean run time
--- a/README.md
+++ b/README.md
@@ -436,6 +436,12 @@ itself.
 
 A `'drain'` event is emitted whenever the `queueSize` reaches `0`.
 
+### Event: `'needsDrain'`
+
+Similar to [`Piscina#needsDrain`](#property-needsdrain-readonly);
+this event is triggered once the total capacity of the pool is exceeded
+by number of tasks enequeued that are pending of execution.
+
 ### Event: `'message'`
 
 A `'message'` event is emitted whenever a message is received from a worker thread.
--- a/README.md
+++ b/README.md
@@ -505,7 +505,7 @@ The current number of tasks waiting to be assigned to a Worker thread.
 
 ### Property: `needsDrain` (readonly)
 
-Boolean value that specifices wether the capacity of the pool has
+Boolean value that specifies whether the capacity of the pool has
 been exceeded by the number of tasks submitted.
 
 This property is helpful to make decisions towards creating backpressure
--- a/src/index.ts
+++ b/src/index.ts
@@ -522,6 +522,7 @@ class ThreadPool {
   runTime : Histogram;
   waitTime : Histogram;
   start : number = performance.now();
+  needsDrain : boolean = false;
   inProcessPendingMessages : boolean = false;
   startingUp : boolean = false;
   workerFailsDuringBootstrap : boolean = false;
@@ -874,8 +875,16 @@ class ThreadPool {
   }
 
   _maybeDrain () {
-    if (this.taskQueue.size === 0 && this.skipQueue.length === 0) {
+    const totalCapacity = this.options.maxQueue + this.pendingCapacity();
+    const totalQueueSize = this.taskQueue.size + this.skipQueue.length;
+
+    if (totalQueueSize === 0) {
       this.publicInterface.emit('drain');
+      this.needsDrain = false;
+    }
+
+    if (totalQueueSize > totalCapacity) {
+      this.needsDrain = true;
     }
   }
 
@@ -1105,6 +1114,10 @@ class Piscina extends EventEmitterAsyncResource {
 
   get duration () : number {
     return performance.now() - this.#pool.start;
+  }
+
+  get needsDrain () : boolean {
+    return this.#pool.needsDrain;
   }
 
   static get isWorkerThread () : boolean {
--- a/src/index.ts
+++ b/src/index.ts
@@ -466,7 +466,7 @@ class WorkerInfo extends AsynchronouslyCreatedResource {
     } catch (err) {
       // This would mostly happen if e.g. message contains unserializable data
       // or transferList is invalid.
-      taskInfo.done(err);
+      taskInfo.done(<Error>err);
       return;
     }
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -507,6 +507,14 @@ An Array of the `Worker` instances used by this pool.
 
 The current number of tasks waiting to be assigned to a Worker thread.
 
+### Property: `needsDrain` (readonly)
+
+Boolean value that specifices wether the capacity of the pool has
```
RafaelGSS:
```suggestion
Boolean value that specifies whether the capacity of the pool has
```

src/index.ts
```diff
@@ -874,8 +875,16 @@ class ThreadPool {
...
+    const totalCapacity = this.options.maxQueue + this.pendingCapacity();
+    const totalQueueSize = this.taskQueue.size + this.skipQueue.length;
+
+    if (totalQueueSize === 0) {
       this.publicInterface.emit('drain');
+      this.needsDrain = false;
+    }
+
+    if (totalQueueSize > totalCapacity) {
+      this.needsDrain = true;
```
RafaelGSS, line 892: shouldn't it be an event emitter instead?

src/index.ts
```diff
@@ -874,8 +875,16 @@ class ThreadPool {
   }
 
   _maybeDrain () {
-    if (this.taskQueue.size === 0 && this.skipQueue.length === 0) {
+    const totalCapacity = this.options.maxQueue + this.pendingCapacity();
+    const totalQueueSize = this.taskQueue.size + this.skipQueue.length;
+
+    if (totalQueueSize === 0) {
       this.publicInterface.emit('drain');
+      this.needsDrain = false;
```
RafaelGSS, line 887: why not `this.pool.needsDrain`?

test/post-task.ts
```diff
@@ -96,18 +96,40 @@ test('postTask() validates abortSignal', async ({ rejects }) => {
...
+  const pool = new Piscina({
+    filename: resolve(__dirname, 'fixtures/eval.js')
+  });
+
+  let drained = false;
+  pool.on('drain', () => {
+    drained = true;
+  });
+
+  const promises = Promise.all([pool.runTask('123'), pool.runTask('123'), pool.runTask('123')]);
```
RafaelGSS: what's the maximum capacity? 3 seems too low.

src/index.ts
metcoder95: Do you mean to hold it at the `publicInterface` instead?

src/index.ts
metcoder95: The original issue seems to aim for a `property` so they can use it to make decisions further. And, you're right, an event can also provide good ergonomics so system metrics or further actions can be taken.

As args of the event, we can pass `totalCapacity`, and `totalQueueSize` at the moment the event takes place.

wdyt?

test/post-task.ts
metcoder95: Hmm, you are right, didn't notice this, I thought I set a low number of `maxThreads` to test it out.

Although, it should be the sum of the workers pending to be instantiated multiplied by `concurrentTasksPerWorker` (defaults to 1), plus `maxQueue`.

The pending capacity seems heavily influenced by the workers spawned at a given time (if using defaults, the max number of threads is the CPU count * `1.5`).

In theory, here it should be up, as my machine is 8 chores gives a total of ~12 threads, meaning that at least 9 of thems should be pending of being instantiated.

Let me go back to this and see what might be missing from my side 🤔

src/index.ts
metcoder95: Added in b2952f3, let me know what you thinkg

README.md
```diff
@@ -507,6 +513,14 @@ An Array of the `Worker` instances used by this pool.
 
 The current number of tasks waiting to be assigned to a Worker thread.
 
+### Property: `needsDrain` (readonly)
```
RafaelGSS, line 512: I think we don't need it anymore

src/index.ts
RafaelGSS: I think the needsDrain = false should be executed when ` (totalQueueSize >= totalCapacity) == false`. If you reset only when is 0, the queue will not be full, and `needsDrain` will show true.

test/post-task.ts
```diff
@@ -96,19 +96,48 @@ test('postTask() validates abortSignal', async ({ rejects }) => {
...
   });
 
-  await Promise.all([pool.runTask('123'), pool.runTask('123')]);
+  await Promise.all([pool.run('123'), pool.run('123')]);
 
   ok(drained);
+  notOk(needsDrain);
+});
+
+test('Piscina exposes/emits needsDrain to true when capacity is exceeded', { only: true }, async ({ ok }) => {
```
RafaelGSS:
```suggestion
test('Piscina exposes/emits needsDrain to true when capacity is exceeded', async ({ ok }) => {
```

README.md
metcoder95: This is what will enable features like the one described in the issue:

```js
if (pool.needDrain) {
  await EE.once(pool, 'drain')
}
pool.runTask()
```

Otherwise, tricks will be required to use a flag on userland whenever the pool needs to be drained

RafaelGSS: but, wouldn't that be the case if the feature just subscribes to pool.on('drain')? am I missing something?

ronag: Yes you are missing something. The drain event only means that needaDrain goes from true to false, you still need the property for the other way around.

src/index.ts
ronag: Please check how writable does. I think it sets needsSrain to false before emitting drain.

README.md
RafaelGSS: Oh, right. Maybe the terminology isn't clear enough.

RafaelGSS (dismissed):
LGTM with @ronag suggestion

src/index.ts
```diff
@@ -874,9 +880,18 @@ class ThreadPool {
...
+    const totalQueueSize = this.taskQueue.size + this.skipQueue.length;
+
+    if (totalQueueSize === 0) {
+      this.needsDrain = false;
       this.publicInterface.emit('drain');
     }
+
+    if (totalQueueSize >= totalCapacity) {
+      this.publicInterface.emit('needsDrain');
+      this.needsDrain = true;
```
```suggestion
      this.needsDrain = true;
      this.publicInterface.emit('needsDrain');
```

ronag: Not sure what this event is for?

metcoder95: Meant to extend the API only, so you can either listen or check the variable

test/post-task.ts
```diff
@@ -96,19 +96,48 @@ test('postTask() validates abortSignal', async ({ rejects }) => {
...
+  let triggered = false;
+  let drained = false;
+  pool.once('drain', () => {
+    drained = true;
+  });
+  pool.once('needsDrain', () => {
+    triggered = true;
+  });
+
+  pool.run('123');
```
RafaelGSS, line 133: Is it reliable? Is there a chance for the first '123' to be executed before the last `pool.run('123')` so pool.once('needsDrain') won't be called?

metcoder95: It should, as we are limiting the threads to 1, so we just paid that overhead once. Then by the fourth one, the first is about or already finished, so the `needsDrain` is already triggered by then, the same as `drain`.

It can sound tricky, but it works 😅 

Do you have a suggestion of how to improve it?

RafaelGSS (approved):

PR #374, opened 2023-07-14
fix: use CJS imports

SimenB:
The `eventemitter-asyncresource` import only works due to using `esModuleInterop`. However, using this flag forces all downstream consumers to also use it, polluting the entire tree.

FWIW, this is the error I get attempting to use Piscina in a project that does not set the interop option.

image

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,12 +1,12 @@
 import { Worker, MessageChannel, MessagePort, receiveMessageOnPort } from 'worker_threads';
 import { once } from 'events';
-import EventEmitterAsyncResource from 'eventemitter-asyncresource';
+import EventEmitterAsyncResource = require('eventemitter-asyncresource');
 import { AsyncResource } from 'async_hooks';
 import { cpus } from 'os';
 import { fileURLToPath, URL } from 'url';
 import { resolve } from 'path';
 import { inspect, types } from 'util';
-import assert from 'assert';
+import assert = require('assert');
 import { Histogram, build } from 'hdr-histogram-js';
 import { performance } from 'perf_hooks';
 import hdrobj from 'hdr-histogram-percentiles-obj';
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -1,12 +1,12 @@
...
 import { once } from 'events';
-import EventEmitterAsyncResource from 'eventemitter-asyncresource';
+import EventEmitterAsyncResource = require('eventemitter-asyncresource');
 import { AsyncResource } from 'async_hooks';
 import { cpus } from 'os';
 import { fileURLToPath, URL } from 'url';
 import { resolve } from 'path';
 import { inspect, types } from 'util';
-import assert from 'assert';
+import assert = require('assert');
```
metcoder95: `assert` also gives you the same error?

src/index.ts
```diff
@@ -1,12 +1,12 @@
 import { Worker, MessageChannel, MessagePort, receiveMessageOnPort } from 'worker_threads';
 import { once } from 'events';
-import EventEmitterAsyncResource from 'eventemitter-asyncresource';
+import EventEmitterAsyncResource = require('eventemitter-asyncresource');
```
metcoder95, line 3: I believe we should group them and companion them with a note of why we are importing them in this manner.

src/index.ts
SimenB: yeah. can do `import * as assert from 'assert'`, but `assert` is used as a function, so that breaks. `import { strict as assert } from 'assert'` possibly?

SimenB: or rather, this doesn't give an error in _my_ code, as the import is removed from the `d.ts` file. It just gives an error if turning off `esModuleInterop`. So the change can be removed from this PR and still solve my issue

src/index.ts
SimenB: Can do. But note that if we set `esModuleInterop` to `false`, it's a TS error to not do this

src/index.ts
metcoder95: Let's revert this one to scope only `eventemitter-asyncresource` for the PR. We can visit the `assert` in #375, wdyt?

metcoder95 (approved):

PR #375, opened 2023-07-14
fix: disable `esModuleInterop`

SimenB:
#374, but going as far as disabling the option in this repo to avoid regressions.

I'm having some issues getting tests to run with this

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,12 +1,12 @@
 import { Worker, MessageChannel, MessagePort, receiveMessageOnPort } from 'worker_threads';
 import { once } from 'events';
-import EventEmitterAsyncResource from 'eventemitter-asyncresource';
+import EventEmitterAsyncResource = require('eventemitter-asyncresource');
 import { AsyncResource } from 'async_hooks';
 import { cpus } from 'os';
 import { fileURLToPath, URL } from 'url';
 import { resolve } from 'path';
 import { inspect, types } from 'util';
-import assert from 'assert';
+import assert = require('assert');
 import { Histogram, build } from 'hdr-histogram-js';
 import { performance } from 'perf_hooks';
 import hdrobj from 'hdr-histogram-percentiles-obj';
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,7 +5,7 @@ import { availableParallelism } from 'os';
 import { fileURLToPath, URL } from 'url';
 import { resolve } from 'path';
 import { inspect, types } from 'util';
-import assert = require('assert');
+import { strict as assert } from 'assert';
 import { Histogram, RecordableHistogram, createHistogram, performance } from 'perf_hooks';
 import {
   READY,
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,15 +1,12 @@
 import { Worker, MessageChannel, MessagePort, receiveMessageOnPort } from 'worker_threads';
-import { once } from 'events';
-import EventEmitterAsyncResource = require('eventemitter-asyncresource');
+import { once, EventEmitterAsyncResource } from 'events';
 import { AsyncResource } from 'async_hooks';
-import { cpus } from 'os';
+import { availableParallelism } from 'os';
 import { fileURLToPath, URL } from 'url';
 import { resolve } from 'path';
 import { inspect, types } from 'util';
 import assert = require('assert');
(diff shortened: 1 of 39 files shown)


test/abort-task.ts
```diff
@@ -1,6 +1,6 @@
 import { AbortController } from 'abort-controller';
 import { EventEmitter } from 'events';
-import Piscina from '..';
+import Piscina = require('..');
```
metcoder95: Let's tackle this differently, by exporting `Piscina` in different manners by exporting it as `default`, and `named export`.

Otherwise, this can potentially break the current user's expectations if `esModuleInterop` is set to `false`. Wdyt?

metcoder95 (commented):
Thanks a lot for the effort, let me know what you think about the comment 🙂

test/abort-task.ts
SimenB: does the default export created by TS work as an actual default export in Node? The current one does as it's just `module.exports =`

src/index.ts
```diff
@@ -1,12 +1,12 @@
...
 import { once } from 'events';
-import EventEmitterAsyncResource from 'eventemitter-asyncresource';
+import EventEmitterAsyncResource = require('eventemitter-asyncresource');
 import { AsyncResource } from 'async_hooks';
 import { cpus } from 'os';
 import { fileURLToPath, URL } from 'url';
 import { resolve } from 'path';
 import { inspect, types } from 'util';
-import assert from 'assert';
+import assert = require('assert');
```
metcoder95: Maybe we can change the way this is imported as discussed in #374 ?

test/abort-task.ts
metcoder95: If exported by default, it will export it as `module.exports.default = Piscina`, if by named export, it translates to `module.Piscina = Piscina`. If just used as `export Piscina`, it just does `module.exports = Piscina`.

Exporting it as default, and named will enable imports like:

- `import { Piscina } from 'piscina'`
- `import Piscina, { SomeType } from 'piscina'`

SimenB:
Completely forgot about this - can take a look at reviving tomorrow

test/abort-task.ts
SimenB: seems #517 did this 👍

SimenB:
@metcoder95 due to https://github.com/piscinajs/piscina/pull/517, this PR simply needs to flip the flag and fix the `assert` import 👍

src/index.ts
```diff
@@ -5,7 +5,7 @@ import { availableParallelism } from 'os';
 import { fileURLToPath, URL } from 'url';
 import { resolve } from 'path';
 import { inspect, types } from 'util';
-import assert from 'assert';
+import { strict as assert } from 'assert';
```
SimenB, line 8: I doubt the change to use `strict` is visible to consumers, but `assert` is used as a function, so we need to tweak this import

metcoder95 (approved):
Thanks! LGTM 👍

metcoder95:
Sorry, finally made the tests works as expected; overall it seems there's an issue with the exports 🤔

SimenB:
Oh, interesting! I'll take a new look tomorrow 😅

SimenB:
Right, the problem is that the types file do not reflect what is actually used at runtime by TS - types point to `main.d.ts`, but `require` loads a file which is just `export =` for CJS compat. This only works with `esModuleInterop` active. Not sure how to best handle this if you do want `const Piscina = require('piscina');` work. Might require a major that only has named exports?

metcoder95:
I've been thinking about refactoring the way the types are overall exported, as currently everything happens with the `Piscina` class, which is practical but it seems that it comes with the tradeoffs of constrains at the moment of importing it that requires to do some tricks.

If you are willing to, I'm more than happy support you on a PR that refactors the way types are exported. Meanwhile we are able to do 
```typescript
const Piscina = require('piscina');
// or
const { Piscina } = require('piscina');

// and ofc
import { Piscina, HelperPiscinaTypes } from 'piscina';
import Piscina, { HelperPiscinaMethods } from 'piscina'; 
```

SimenB:
not stale, just low on OSS time 😅 It's on my list!

SimenB:
On vacation, will get back to this (hopefully in August)

PR #396, opened 2023-08-31
feat: Add `Piscina#close` API

### LINKED ISSUE #139: Graceful Shutdowns
I don't see anything to the affect of, "Method `foo` will signal to Piscina that the application is shutting down. Piscina will then clear the queue, wait for running tasks to finish, and emit an event when it is drained". Does this exist in Piscina?

andersonjoseph:
Hi everyone!

This PR adds a new method called `close()`. It allows you to stop all the workers in a pool gracefully (closes: #139)

Please let me know if you have any feedback or suggestions for improvement. I'm happy to make any changes if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,6 +420,23 @@ Stops all Workers and rejects all `Promise`s for pending tasks.
 
 This returns a `Promise` that is fulfilled once all threads have stopped.
 
+### Method: `close([options])`
+
+* `options`:
+  * `force`: A `boolean` value that indicates whether to abort all tasks that 
+  are enqueued but not started yet. The default is `false`.
+  * `timeout`: A `number` time (in milliseconds) to wait for the pool to 
+  complete all in-flight tasks. The default is `3000`
+
+Stops all Workers gracefully.
+
+This returns a `Promise` that is fulfilled once all tasks that were started 
+have completed and all threads have stopped.
+
+This method is similar to `destroy()`, but with the difference that `close()` 
+will wait for the worker tasks to finish, while `destroy()` 
+will abort them immediately.
+
 ### Event: `'error'`
 
 An `'error'` event is emitted by instances of this class when:
--- a/README.md
+++ b/README.md
@@ -357,7 +357,7 @@ This class extends [`EventEmitter`][] from Node.js.
     Defaults to `true`. (This option is only supported on Node.js 12.19+ and
     all Node.js versions higher than 14.6.0).
   * `closeTimeout`: (`number`) An optional time (in milliseconds) to wait for the pool to 
-  complete all in-flight tasks when `close()` is called. The default is `3000`
+  complete all in-flight tasks when `close()` is called. The default is `30000`
 
 Use caution when setting resource limits. Setting limits that are too low may
 result in the `Piscina` worker threads being unusable.
--- a/README.md
+++ b/README.md
@@ -356,6 +356,8 @@ This class extends [`EventEmitter`][] from Node.js.
     `fs.close()`, and will close them automatically when the Worker exits.
     Defaults to `true`. (This option is only supported on Node.js 12.19+ and
     all Node.js versions higher than 14.6.0).
+  * `closeTimeout`: (`number`) An optional time (in milliseconds) to wait for the pool to 
+  complete all in-flight tasks when `close()` is called. The default is `3000`
 
 Use caution when setting resource limits. Setting limits that are too low may
 result in the `Piscina` worker threads being unusable.
@@ -425,8 +427,6 @@ This returns a `Promise` that is fulfilled once all threads have stopped.
 * `options`:
   * `force`: A `boolean` value that indicates whether to abort all tasks that 
   are enqueued but not started yet. The default is `false`.
-  * `timeout`: A `number` time (in milliseconds) to wait for the pool to 
-  complete all in-flight tasks. The default is `3000`
 
 Stops all Workers gracefully.
 
--- a/README.md
+++ b/README.md
@@ -457,7 +457,7 @@ A `'drain'` event is emitted whenever the `queueSize` reaches `0`.
 
 Similar to [`Piscina#needsDrain`](#property-needsdrain-readonly);
 this event is triggered once the total capacity of the pool is exceeded
-by number of tasks enequeued that are pending of execution.
+by number of tasks enqueued that are pending of execution.
 
 ### Event: `'message'`
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -149,7 +149,7 @@ const kDefaultOptions : FilledOptions = {
   taskQueue: new ArrayTaskQueue(),
   niceIncrement: 0,
   trackUnmanagedFds: true,
-  closeTimeout: 3000
+  closeTimeout: 30000
 };
 
 interface RunOptions {
--- a/src/index.ts
+++ b/src/index.ts
@@ -1001,6 +1001,7 @@ class ThreadPool {
       this.publicInterface.emit('error', error);
     } finally {
       await this.destroy();
+      this.publicInterface.emit('close');
       this.closingUp = false;
     }
   }
--- a/src/index.ts
+++ b/src/index.ts
@@ -1000,9 +1000,8 @@ class ThreadPool {
       this.publicInterface.emit('error', error);
     } finally {
       await this.destroy();
+      this.closingUp = false;
     }
-
-    this.closingUp = false;
   }
 }
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -120,6 +120,7 @@ interface Options {
   taskQueue? : TaskQueue,
   niceIncrement? : number,
   trackUnmanagedFds? : boolean,
+  closeTimeout?: number
 }
 
 interface FilledOptions extends Options {
@@ -132,7 +133,8 @@ interface FilledOptions extends Options {
   concurrentTasksPerWorker : number,
   useAtomics: boolean,
   taskQueue : TaskQueue,
-  niceIncrement : number
+  niceIncrement : number,
+  closeTimeout : number
 }
 
 const kDefaultOptions : FilledOptions = {
@@ -146,7 +148,8 @@ const kDefaultOptions : FilledOptions = {
   useAtomics: true,
   taskQueue: new ArrayTaskQueue(),
   niceIncrement: 0,
-  trackUnmanagedFds: true
+  trackUnmanagedFds: true,
+  closeTimeout: 3000
 };
 
 interface RunOptions {
@@ -172,12 +175,10 @@ const kDefaultRunOptions : FilledRunOptions = {
 
 interface CloseOptions {
   force?: boolean,
-  timeout?: number
 }
 
 const kDefaultCloseOptions : Required<CloseOptions> = {
-  force: false,
-  timeout: 3000
+  force: false
 };
 
 class DirectlyTransferable implements Transferable {
@@ -994,7 +995,7 @@ class ThreadPool {
     try {
       await Promise.race([
         onPoolFlushed(),
-        throwOnTimeOut(options.timeout)
+        throwOnTimeOut(this.options.closeTimeout)
       ]);
     } catch (error) {
       this.publicInterface.emit('error', error);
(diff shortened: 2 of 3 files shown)


src/index.ts
```diff
@@ -913,6 +934,60 @@ class ThreadPool {
 
     await Promise.all(exitEvents);
   }
+
+  async close (options : CloseOptions = kDefaultCloseOptions) {
+    this.closingUp = true;
+
+    if (options.force) {
```
metcoder95, line 947: I think this is equivalent to doing `pool.destroy`, not so sure if really needed. What do you think?

src/index.ts
```diff
@@ -913,6 +934,60 @@ class ThreadPool {
...
+        const taskInfo : TaskInfo = this.skipQueue.shift() as TaskInfo;
+        if (taskInfo.workerInfo === null) {
+          taskInfo.done(new AbortError());
+        } else {
+          this.skipQueue.push(taskInfo);
+        }
+      }
+
+      const taskQueueLength = this.taskQueue.size;
+      for (let i = 0; i < taskQueueLength; i++) {
```
metcoder95, line 959: The equivalent will be:

```suggestion
	for (const taskInfo of skipQueue) {
```

src/index.ts
```diff
@@ -913,6 +934,60 @@ class ThreadPool {
...
+        const taskInfo : TaskInfo = this.taskQueue.shift() as TaskInfo;
+        if (taskInfo.workerInfo === null) {
+          taskInfo.done(new AbortError());
+        } else {
+          this.taskQueue.push(taskInfo);
+        }
```
metcoder95, line 965: I'm not fully convinced that aborting them is the right path to choose, I'd prefer to find a way to flush them before a timeout.
My expectation is similar to what happens when you run `httpServer.close`, current requests are being flushed out, but no new ones are being accepted.

src/index.ts
```diff
@@ -913,6 +934,60 @@ class ThreadPool {
...
+      for (const workerInfo of this.workers) {
+        removeWorkerIfIsDone(workerInfo);
+        resolveIfWorkersIsEmpty();
+
+        workerInfo.port.on('message', () => {
+          removeWorkerIfIsDone(workerInfo);
+          resolveIfWorkersIsEmpty();
+        });
+      }
+    });
```
metcoder95: It will be more robust if we combine `close` with `destroy`; meaning that we call `destroy` after we flush the queue or ran into a timeout

test/pool-close.ts
```diff
@@ -0,0 +1,70 @@
...
+    const pool = new Piscina({ filename: resolve(__dirname, 'fixtures/eval.js') });
+    await pool.close();
+    t.pass('pool closed successfully');
+  });
+
+  t.test('queued tasks waits for all tasks to complete', async (t) => {
+    const pool = new Piscina({ filename: resolve(__dirname, 'fixtures/eval.js'), maxThreads: 1 });
+
+    const task1 = pool.run('1 + 1');
+    const task2 = pool.run('2 + 2');
```
metcoder95: I'd recommend using more complex tasks, like a `setTimeout` with more than a few milliseconds to test out it truly flushes the queue within a time window.

metcoder95 (commented):
Hey! Sorry for the fast review but it kept my attention as I was working on the same feature 😅

src/index.ts
andersonjoseph: the `force` option is a bit different to `pool.destroy`

using `close` with `force: true` will wait for the running tasks to complete before terminating the workers, while `pool.destroy` will terminate all of them without waiting.

here's a small example: 

```ts
  // using destroy()

  const worker = new Piscina({
    filename: resolve(__dirname, 'fixtures/sleep.js'),
    maxThreads: 2,
  });

  setTimeout(() => {
    worker.destroy();
  }, 100);

  const result = await Promise.allSettled([
    worker.run({time: 10000, a: 1}),
    worker.run({time: 1000, a: 1}),
    worker.run({time: 1000, a: 1}),
  ])

  console.log(result); // all promises will be rejected
```

```ts
  // using close({force: true})

  const worker = new Piscina({
    filename: resolve(__dirname, 'fixtures/sleep.js'),
    maxThreads: 2,
  });

  setTimeout(() => {
    worker.close({force: true});
  }, 100);

  const x = await Promise.allSettled([
    worker.run({time: 10000, a: 1}),
    worker.run({time: 1000, a: 1}),
    worker.run({time: 1000, a: 1}),
  ])

  console.log(x); // only one will be rejected because the first two were already running
```

src/index.ts
andersonjoseph: yeah I wasn't sure to using `AbortError` here, I tried to use a different approach where I flushed the queue, however the `run` promises were fulfilled with `null` (which didn't convinced me, since you expect a task to complete with your expected result or fail in the other case)

if you could point me in the right direction to implement something similar to `httpServer.close` having the above in mind, it would be great

src/index.ts
metcoder95: And without the `force: true`?
If I'm using the API, I would expect:

1. Calling `close({ force: true })`, will shut down the pool right away,  aborting the current tasks in-flight and the remaining in the queue waiting to be processed.
2. Calling `close()`, will wait until all the tasks are being processed before shutting down the pool (maybe add a timeout to not avoid hanging threads).

One seems quite analogous to `destroy` as its basically what it does

src/index.ts
metcoder95: This is the rough implementation I was doing:

```ts
  async close (): Promise<void> {
    if (this.closing || this.closed) return;

    this.closing = true;

    // Flush high-prio enqueued tasks
    const result = await Promise.race([
      flushPool.call(this),
      settimeout(this.options.closeTimeout ?? 60000) // default to 1 minute
    ]);

    const err = !result ? new Error('closed timed out') : null;

    await this.destroy();
    this.closed = true;
    this.closing = false;

    if (err) this.publicInterface.emit('error', err);
    this.publicInterface.emit('close', err);

    // TODO: Flush is hanging, it seems it is not getting updates
    // from the workers to account for the number of current tasks in flight
    async function flushPool (this: ThreadPool): Promise<boolean> {
      while (this._getTaskCountInFlight() !== 0) {
        while (this.skipQueue.length !== 0) {
          const taskInfo = <TaskInfo> this.skipQueue.shift();
          const candidate = this.workers.findAvailable();

          if (!candidate || this.needsDrain) {
            await once(this.publicInterface, 'drain');
            this.skipQueue.push(taskInfo);
            continue;
          }

          const now = performance.now();
          this.waitTime.recordValue(now - taskInfo.created);
          taskInfo.started = now;
          candidate.postTask(taskInfo);
          this._maybeDrain();
        }

        // Flush lower prio tasks
        while (this.taskQueue.size !== 0) {
          const taskInfo : TaskInfo = this.taskQueue.shift() as TaskInfo;
          const candidate = this.workers.findAvailable();

          if (!candidate || this.needsDrain) {
            await once(this.publicInterface, 'drain');
            this.taskQueue.push(taskInfo);
            continue;
          }

          const now = performance.now();
          this.waitTime.recordValue(now - taskInfo.created);
          taskInfo.started = now;
          candidate.postTask(taskInfo);
          this._processPendingMessages();
          this._maybeDrain();
        }

	    this._processPendingMessages();
      }
      return true;
    }
  }
```

The important piece is the `flushPool` which is the one seeking to flush all the current and pending tasks within the queue and pool until all.

src/index.ts
andersonjoseph: Without `force: true`, `close()` will wait for all the current running tasks **and** the enqueued tasks to complete before closing all workers. (I was also thinking about adding a timeout to the options and running destroy when the timeout expires)

I took this approach based on [this comment from Matteo](https://github.com/piscinajs/piscina/issues/139#issuecomment-883160289), but if your implementation is closer to the right path, I have no problem in closing this PR and helping with yours in any way I can :)

metcoder95: Let's continue with your work, it's always nice to have new people contributing to the project 😄
Hmm, ok I see. Well, the thing is that `destroy` already does that, but it's true that it goes beyond as the current ones running are aborted. Then, wdyt if we set the following specs as a baseline for this work:

1. `Piscina#close` should stop receiving any new task, and flush the ones that are currently enqueued and in-flight (currently being processed by a worker)
2. If `Piscina#close` is called with `{ force: true }`, we abort all the `taskQueue` and `skipQueue`, but wait until the worker finishes its current tasks assigned.
3. Calling `Piscina#close`, calls `Piscina#destroy` underneath to close all the workers alive.
4. Through `new Piscina({ closeTimeout: 30000 })` we accept a timeout to use when calling close
  - If timeout is exceeded, Piscina calls `Piscina#destroy` immediately, destroying all workers and aborting all the tasks
  - The default is `30000` (we need to document this as well)

Wdyt?

andersonjoseph: I made some updates to the code considering those specs, also added a couple of tests for `timeout` cases

let me know what you think 👌

README.md
```diff
@@ -420,6 +420,23 @@ Stops all Workers and rejects all `Promise`s for pending tasks.
 
 This returns a `Promise` that is fulfilled once all threads have stopped.
 
+### Method: `close([options])`
+
+* `options`:
+  * `force`: A `boolean` value that indicates whether to abort all tasks that 
+  are enqueued but not started yet. The default is `false`.
+  * `timeout`: A `number` time (in milliseconds) to wait for the pool to
```
metcoder95: I believe that it has better ergonomics if set when instantiating `Piscina`, so it looks like `new Piscina({ closeTimeout: 40000 })` wdyt?  🤔

src/index.ts
```diff
@@ -913,6 +939,71 @@ class ThreadPool {
...
+        throwOnTimeOut(options.timeout)
+      ]);
+    } catch (error) {
+      this.publicInterface.emit('error', error);
+    } finally {
+      await this.destroy();
+    }
+
+    this.closingUp = false;
+  }
```
metcoder95: `nit`

```suggestion
    } finally {
      await this.destroy();
      this.closingUp = false;
    }
  }
```

src/index.ts
```diff
@@ -913,6 +939,71 @@ class ThreadPool {
...
+    const throwOnTimeOut = async (timeout: number) => {
+      await sleep(timeout);
+      throw Errors.CloseTimeout();
+    };
+
+    try {
+      await Promise.race([
+        onPoolFlushed(),
+        throwOnTimeOut(options.timeout)
+      ]);
```
metcoder95, line 999: We might want to also emit a `close` event once everything is fully closed, before calling destroy. Wdyt?
```suggestion
      ]);
      this.publicInterface.emit('close')
```

metcoder95 (commented):
Great work! Minor comments left 👍

src/index.ts
andersonjoseph: sounds good! quick question: should we emit the `close` event even on timeout?

metcoder95: Yeah, in theory, although we fall into timeout we "_exhausted_" the pool from all in-flight tasks, as well as enqueued 👍

src/index.ts
```diff
@@ -562,6 +577,9 @@ class ThreadPool {
   }
 
   _ensureMinimumWorkers () : void {
+    if (this.closingUp) {
```
metcoder95, line 580: We need to do the same for `runTasks`, we should not accept incoming `messages`.
In fact, if we are not receiving more incoming messages, no new workers will be needed to be made.

andersonjoseph: I'm already doing something similar for `runTasks` [here](https://github.com/piscinajs/piscina/pull/396/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R785-R794):

```ts
    let signal: AbortSignalAny | null;
    if (this.closingUp) {
      const closingUpAbortController = new AbortController();
      closingUpAbortController.abort('queue is closing up');

      signal = closingUpAbortController.signal;
    } else {
      signal = options.signal ?? null;
    }

```

basically adding an aborted signal on task creation. [Here's the test for this behaviour](https://github.com/piscinajs/piscina/pull/396/files#diff-70a06c525481045dcce4f42869e7009bd336dc29428384b238b6c3d2d564532bR28-R37)

metcoder95: LGTM 👍

src/index.ts
```diff
@@ -145,7 +148,8 @@ const kDefaultOptions : FilledOptions = {
   useAtomics: true,
   taskQueue: new ArrayTaskQueue(),
   niceIncrement: 0,
-  trackUnmanagedFds: true
+  trackUnmanagedFds: true,
+  closeTimeout: 3000
```
metcoder95: 3s seems too low. Let's be gentle and start with 30s, wdyt?

metcoder95 (commented):
Nice work 🚀 
Small `nit` and we are ready to go

src/index.ts
andersonjoseph: sounds good! change pushed

metcoder95 (approved):

PR #429, opened 2023-10-20
chore(release): 4.2.0

metcoder95:


alan-agius4:
Would appreciate if  https://github.com/piscinajs/piscina/pull/430 is also included in the release

PR #433, opened 2023-10-23
fix: migrate to EventEmitterAsyncResource from core

### LINKED ISSUE #426: Piscina and @types/node > 20.8.0
We are experiencing some issues with @types/node > 20.8.0 and piscina.
Piscina depends on the module `eventemitter-asyncresource` which has not been updated within 3 years now.

If I try to upgrade to @types/node version 20.8.6, during the build phase this error is raised:
```
Error: node_modules/eventemitter-asyncresource/dist/src/index.d.ts(20,15): error TS2417: Class static side 'typeof EventEmitterAsyncResource' incorrectly extends base class static side 'typeof EventEmitter'.
  Types of property 'EventEmitterAsyncResource' are incompatible.
    Types of construct signatures are incompatible.
      Type 'new (options?: string | Options) => EventEmitterAsyncResource' is not assignable to type 'new (options?: EventEmitterAsyncResourceOptions) => EventEmitterAsyncResource'.
        Construct signature return types 'EventEmitterAsyncResource' and 'EventEmitterAsyncResource' are incompatible.
          The types of 'asyncId' are incompatible between these types.
            Type '() => number' is not assignable to type 'number'.
Error: Process completed with exit code 2.
```

I'm using `typescript@5.1.6` and this is my tsconfig:
```
{
  "compilerOptions": {
    "module": "commonjs",
    "esModuleInterop": true,
    "allowJs": true,
    "allowSyntheticDefaultImports": true,
    "target": "es2017",
    "noImplicitAny": true,
    "moduleResolution": "node16",
    "resolveJsonModule": true,
    "sourceMap": true,
    "outDir": "dist"
  },
  "include": [
    "src/**/*"
  ]
}
```

I hope these informations are enough, otherwise I'll try to create a minimal repro later today.

groozin:
Fixes: #426

- Updated the types definitions to include `EventEmitterAsyncResource`
- Removed external package for `EventEmitterAsyncResource`
- Updated typescript to v5 - needed for the type definitions 
- Fixed type errors after upgrade to ts v5

Note: I guess this loses the compatibility with node versions prior to inclusion of `EventEmitterAsyncResource`

### DIFF
--- a/src/worker.ts
+++ b/src/worker.ts
@@ -172,7 +172,7 @@ function onMessage (
         result: null,
         // It may be worth taking a look at the error cloning algorithm we
         // use in Node.js core here, it's quite a bit more flexible
-        error
+        error: error instanceof Error ? error : null
       };
     }
     currentTasks--;
--- a/package.json
+++ b/package.json
@@ -35,8 +35,8 @@
   "license": "MIT",
   "devDependencies": {
     "@types/node": "^20.8.0",
-    "@typescript-eslint/eslint-plugin": "^4.22.0",
-    "@typescript-eslint/parser": "^4.22.0",
+    "@typescript-eslint/eslint-plugin": "^6.9.0",
+    "@typescript-eslint/parser": "^6.9.0",
     "abort-controller": "^3.0.0",
     "concat-stream": "^2.0.0",
     "gen-esm-wrapper": "^1.1.1",
@@ -44,7 +44,7 @@
     "standardx": "^7.0.0",
     "tap": "^16.3.7",
     "ts-node": "^9.1.1",
-    "typescript": "5.2.x"
+    "typescript": "5.1.6"
   },
   "dependencies": {
     "hdr-histogram-js": "^2.0.1",
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
   ],
   "license": "MIT",
   "devDependencies": {
-    "@types/node": "^20.3.3",
+    "@types/node": "^20.8.0",
     "@typescript-eslint/eslint-plugin": "^4.22.0",
     "@typescript-eslint/parser": "^4.22.0",
     "abort-controller": "^3.0.0",
@@ -44,10 +44,9 @@
     "standardx": "^7.0.0",
     "tap": "^16.3.7",
     "ts-node": "^9.1.1",
-    "typescript": "4.3.x"
+    "typescript": "5.2.x"
   },
   "dependencies": {
-    "eventemitter-asyncresource": "^1.0.0",
     "hdr-histogram-js": "^2.0.1",
     "hdr-histogram-percentiles-obj": "^3.0.0"
   },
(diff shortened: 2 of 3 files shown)


simoneb (approved):
LGTM

src/worker.ts
```diff
@@ -172,7 +172,7 @@ function onMessage (
         result: null,
         // It may be worth taking a look at the error cloning algorithm we
         // use in Node.js core here, it's quite a bit more flexible
-        error
+        error: error instanceof Error ? error : null
```
metcoder95, line 175: Why this change?

metcoder95 (commented):
LGTM, but lint is failing; can you check?

>I guess this loses the compatibility with node versions prior to inclusion of EventEmitterAsyncResource

That's fine, `EventEmitterAsyncResource` was added on v16, which is the minimum version we support 👍

src/worker.ts
groozin: In recent versions of typescript the assignment of `unknown` (error in this context) is impossible to defined type `Error | null`.
See below the exact error.

```bash
> npm run build

> piscina@4.1.0 build
> tsc && gen-esm-wrapper . dist/esm-wrapper.mjs

src/worker.ts:175:9 - error TS2322: Type 'unknown' is not assignable to type 'Error | null'.

175         error
            ~~~~~

  src/common.ts:28:3
    28   error: Error | null;
         ~~~~~
    The expected type comes from property 'error' which is declared here on type 'ResponseMessage'


Found 1 error in src/worker.ts:175
```

groozin: @metcoder95 checking the linting errors.

metcoder95: LGTM 👍

groozin: OK. So the linting error is:
```bash
/r/piscina/test/test-uncaught-exception-from-handler.ts
  0:0  error  Parsing error: DeprecationError: 'originalKeywordKind' has been deprecated since v5.0.0 and can no longer be used. Use 'identifierToKeywordKind(identifier)' instead
```
And it appears on the very top of, pretty much, every `ts` file.

Now I used the `rg --no-ignore originalKeywordKind` to look for that keyword and it only appears in `node_modules` packages related to typescript and lining. And not a single occurrence in the `src` folder.

My best guess here it is a configuration and compatibility issue between `ts 5.x.x.` and linter. Since this package is using `standardx`, then the underlying `eslint` version is `7` (the oldish `standard` dependency). I would suggest updating the `standardx` dependency once it updates its own dependency. There is a PR for that: https://github.com/standard/standardx/pull/42 - however it is stale for some time.

I tried few different package versions of typescript and linter plugins, also tried to do `npm override` on the `eslint`
package version - but ran into breaking changes issue.

metcoder95: Interesting, good catch 👍 

Does [`skipLibCheck`](https://www.typescriptlang.org/tsconfig#skipLibCheck) can help us for the moment?

groozin: I made some more changes to linter plugin versions and version of typescript (still 5th as major) and that has gotten rid of the linter errors locally.

metcoder95 (approved):

metcoder95:
Thanks for the contribution!

ilteoood:
@metcoder95 may I ask you to create a new release for it?

metcoder95:
I'll reuse the current one 👍

PR #447, opened 2023-11-19
chore: Improve package size

pavelkornev:
Include only relevant files in the bundle, everything else is skipped from the shipped NPM package.

Before: 424 KB
After: 164 KB

`.npmignore` is no longer needed, allowed files are controlled via [files](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#files) option in `package.json`.


metcoder95 (commented):
We would ideally want to also include `test` files to enhance DX (offline mode examples). 
The rest it’s ok.

PR #457, opened 2023-11-28
fix: default minThreads with odd CPU count

Chocobozzz:
If not, Piscina will create and remove the latest worker indefinitely

Because of https://github.com/piscinajs/piscina/blob/current/src/index.ts#L587 and https://github.com/piscinajs/piscina/blob/current/src/index.ts#L762


metcoder95 (approved):
lgtm

PR #482, opened 2024-01-12
feat: use native Node.js histogram support

clydin:
The histogram related dependencies (`hdr-histogram-js`/`hdr-histogram-percentiles-obj`) have been replaced with the native support present in Node.js. This allows the package to have no required dependencies (with one optional for nice support on linux). The public API surface was kept in place to prevent breaking changes for users. This required a small amount of additional code to mainly transform the percentiles to the previous structure.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,17 +1,13 @@
 {
   "name": "piscina",
   "version": "4.2.1",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
   "requires": true,
   "packages": {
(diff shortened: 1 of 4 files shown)


metcoder95 (commented):
Thanks a lot, now we can fully say that `piscina` will be dependency-less!

Can you just check lint, it seems is failing

clydin:
Addressed the lint errors.

metcoder95 (approved):

Issue #452, opened 2023-11-26
Default TaskQueue implementation is slow for a large number of queued tasks


robhogan:
*I'm writing this not as an ask from the maintainers but as a note for users and an opportunity for contributors.*

While benchmarking `piscina` against `jest-worker` for a (very) large number of small tasks, I saw `jest-worker` was several times faster out of the box, using worker threads. After digging in a bit, it turns out that for a large enough number of tasks, `ArrayTaskQueue.prototype.shift()`, within the default FIFO queue , whose [implementation](https://github.com/piscinajs/piscina/blob/24822f4532d128b1bf6347f28f765512529c5863/src/index.ts#L95) is just an O(n) `Array` shift, dominates on the host thread. For me this became apparent at around 100,000 queued tasks.

A lightweight linked list implementation would speed this up algorithmically without much extra complexity in implementation - indeed using a [custom `taskQueue`](https://github.com/piscinajs/piscina#custom-task-queues)  I was able to get parity with `jest-worker` for queues around 500k.

PR to follow (hopefully!), but I wanted to note this anyway for the benefit of anyone seeing (maybe without realising) this issue.

metcoder95:
Hey @robhogan, thanks for your notes; they are in fact very useful.
Overall, it is somehow expected that a common FIFO algorithm for balancing tasks will face some limitations with big tasks numbers, the goal is not make it a one-size-fits-all solution. That's why `CustomTaskQueue` exists.

We have #347 to enhance and have some out-of-the-box algorithms that can fulfills other needs, aside of the `CustomTaskQueue` one.

I'm seeing this one as somehow duplicated of #347, I'll link your comment there for reference.

ronag:
Use nodes fixed_queue implementation?

metcoder95:
As `default` Task Queue?

JaoodxD:
@ronag @metcoder95 I would like to enhance piscina with node's fixed_queue. 
Should I firstly create separate issue or straightly go to work and then make a PR?

metcoder95:
Let's continue on the PR 👍

Issue #347, opened 2023-06-07
More customisable scheduling


ronag:
Currently, piscina only supports depth-first scheduling. It would be nice to be able to add more scheduling strategies based on different types of heuristics.

A bit like https://github.com/poolifier/poolifier but not necessarily as advanced.

metcoder95:
We can start with something small like Round Robin, and its weighted version. And allow, through options, how to assign weights and other moving pieces.

+1 on going for it 👍

PR #487, opened 2024-01-19
feat!: Custom Scheduler

metcoder95:
The PR attempts to start the decomposition of the task scheduler that distributes tasks among queues into a separate module. Seeking to enable custom task scheduling for Piscina.

Relates to #452 & #347

- [x] Documentation
	- [x] Document new `PISCINA_THREAD_ID` env var
- [x] Tests for `BaseScheduler` class
- [x] Tests with Custom Schedulers
- [x] Define an interface for WorkerInfo


src/scheduler.ts
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  delete (_worker: PiscinaWorker): void {
+    throw new Error('delete Method not implemented.');
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  onNewWorker (_worker: PiscinaWorker): void {
+    throw new Error('maybeAvailable Method not implemented.');
```
metcoder95:
```suggestion
    throw new Error('onNewWorker Method not implemented.');
```

src/scheduler.ts
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  add (_worker: PiscinaWorker): void {
+    throw new Error('add Method not implemented.');
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  pick (_task: PiscinaTask): PiscinaWorker | null {
+    throw new Error('findAvailable Method not implemented.');
```
metcoder95:
```suggestion
    throw new Error('pick Method not implemented.');
```

src/scheduler.ts
```diff
@@ -0,0 +1,139 @@
...
+    throw new Error('delete Method not implemented.');
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  onNewWorker (_worker: PiscinaWorker): void {
+    throw new Error('maybeAvailable Method not implemented.');
+  }
+
+  getAvailableCapacity (): number {
+    throw new Error('getCurrentCapacity Method not implemented.');
```
metcoder95:
```suggestion
    throw new Error('getAvailableCapacity Method not implemented.');
```

metcoder95:
cc @RafaelGSS @jasnell @ronag it should be open for review now; happy to start the feedback cycle on this 🙂

src/common.ts
```diff
@@ -68,7 +105,49 @@ export interface Transferable {
...
+      listener();
+    }
+  }
+
+  isReady () : boolean {
+    return this.onreadyListeners === null;
+  }
+
+  onReady (fn : () => void) {
+    if (this.onreadyListeners === null) {
```
RafaelGSS:
```suggestion
    if (this.isReady()) {
```

src/index.ts
```diff
@@ -474,46 +366,46 @@ class WorkerInfo extends AsynchronouslyCreatedResource {
...
-    this.clearIdleTimeout();
+    this.#clearIdleTimeout();
     for (const taskInfo of this.taskInfos.values()) {
       taskInfo.done(Errors.ThreadTermination());
     }
     this.taskInfos.clear();
   }
 
-  clearIdleTimeout () : void {
+  #clearIdleTimeout () : void {
```
RafaelGSS, line 395: Wouldn't it make this change a semver-major? I think you can do it in a separate PR

src/scheduler.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+}
+
+class DefaultTaskScheduler extends TaskScheduler {
+  #pendingItems = new Set<PiscinaWorker>();
+  #readyItems = new Set<PiscinaWorker>();
+  #maximumUsage: number;
+  #onAvailableListeners: ((item: PiscinaWorker) => void)[];
+
+  constructor (maximumUsage: number) {
```
RafaelGSS, line 75: I think we could default it to `os.availableParallelism()` https://nodejs.org/api/os.html#osavailableparallelism

RafaelGSS (commented):
Considering all interface changes, this is semver-major right?

src/scheduler.ts
metcoder95: Let me make an issue out of that 👍

src/index.ts
metcoder95: Agree, but AFAIT we do not expose those internals isn't it?

If so, we should make it; otherwise I believe they can go under a minor, specially as we are not changing the behaviour by default, let me know what you think 🤔

RafaelGSS: I'm not 100% familiar with this API. If you say we don't expose it, that's fine.

metcoder95: Thanks! Yeah, I double checked, we do not expose them, are meant for internal usage 👍

metcoder95:
I'll work Today on sorting out the conflicts, for some reason some changes are missed after that

test/schedulers/custom.ts
```diff
@@ -0,0 +1,286 @@
...
+
+    add (item) {
+      this.#pendingWorkers.add(item);
+      item.onReady(() => {
+        /* istanbul ignore else */
+        if (this.#pendingWorkers.has(item)) {
+          this.#pendingWorkers.delete(item);
+          this.#readyWorkers.add(item);
+          this.onWorkerAvailable(item);
+        }
```
ronag:
```suggestion
        if (this.#pendingWorkers.delete(item)) {
          this.#readyWorkers.add(item);
          this.onWorkerAvailable(item);
        }
```

metcoder95: Done in [0b58d1c](https://github.com/piscinajs/piscina/pull/487/commits/0b58d1c2633033ece1bbcb40bacecfc2c40d63c3)

metcoder95:
Can I have a final review? cc: @RafaelGSS @ronag

README.md
```diff
@@ -709,6 +722,149 @@ options on to the custom `TaskQueue` implementation. (Note that because the
...
+class TaskScheduler {
+  size: number = 0
+
+  // The constructor will receive the max number of concurrent task per worker
+  // as passed when instantiating Piscina
+  constructor (concurrentTasksPerWorker: number) {
+  }
+
+  [Symbol.iterator] (): IterableIterator<PiscinaWorker> {
+    // Meant to enable Piscina iterate over the workers for different purposes
```
ronag, line 723: What's this for? It kind of breaks en encapsulation of workers.

README.md
```diff
@@ -709,6 +722,149 @@ options on to the custom `TaskQueue` implementation. (Note that because the
...
+  }
+
+  add (worker: PiscinaWorker): void {
+    // Will be called when a new worker is created
+    // Workers will be marked as ready after initialization
+    // You can listen to it calling worker.onReady and scheduling a callback
```
ronag, line 729: When are workers created? Shouldn't that be up to the scheduler?

README.md
```diff
@@ -709,6 +722,149 @@ options on to the custom `TaskQueue` implementation. (Note that because the
...
+    // This method is used by Piscina to listen when a new Worker has been marked as ready
+    // after being created.
+  }
+
+  delete (worker: PiscinaWorker): void {
+    // Will be called when a worker must be terminated for different reasons
+  }
+
+  onWorkerAvailable (worker: PiscinaWorker): void {
+    // Will be called whenever a has finished a task and is ready to receive a new one.
```
ronag, line 750: Why no just call add again? and rename pick to shift or something?

README.md
```diff
@@ -709,6 +722,149 @@ options on to the custom `TaskQueue` implementation. (Note that because the
...
+    // It is up to the scheduler decide how it wants to pick a worker for a given task.
+  }
+
+  onAvailable (cb: (worker: PiscinaWorker) => void): void {
+    // This method is used by Piscina to listen when a new Worker has been marked as ready
+    // after being created.
+  }
+
+  delete (worker: PiscinaWorker): void {
+    // Will be called when a worker must be terminated for different reasons
```
ronag: What reasons?

src/common.ts
```diff
@@ -68,7 +105,49 @@ export interface Transferable {
...
+    }
+  }
+
+  isReady () : boolean {
+    return this.onreadyListeners === null;
+  }
+
+  onReady (fn : () => void) {
+    if (this.isReady()) {
+      fn(); // Zalgo is okay here.
```
ronag: Is it? :)

src/common.ts
```diff
@@ -68,7 +105,49 @@ export interface Transferable {
...
+    for (const listener of listeners) {
+      listener();
+    }
+  }
+
+  isReady () : boolean {
+    return this.onreadyListeners === null;
+  }
+
+  onReady (fn : () => void) {
```
ronag: What about failures?

README.md
```diff
@@ -709,6 +722,149 @@ options on to the custom `TaskQueue` implementation. (Note that because the
...
+      this.#maxConcurrent = maxConcurrent;
+      this.#readyWorkers = new Set();
+      this.#pendingWorkers = new Set();
+      this.#onAvailableListeners = [];
+      this.#offset = 0;
+    }
+
+    add (item) {
+      this.#pendingWorkers.add(item);
+      item.onReady(() => {
```
ronag, line 783:
```suggestion
      item.onReady((err) => {
```

ronag (commented):
Any reason for going away from event emitters?

src/common.ts
metcoder95: On that, I'm not sure; was in this way so I just port it. But it might be worth it to see if zalgo is really ok there 🤔

src/common.ts
metcoder95: Hmm, good point; it seems we immediately remove the worker without even calling `onReady`. This might be opaque for the Scheduler. Although I'm not sure how much value it will bring, anyways as the Scheduler just won't account it for its pool

src/common.ts
ronag: just do a queueMicrotask to be sure

src/common.ts
ronag: Then it would e.g. forever stay in the pendingWorkers set, no?

README.md
metcoder95: I think I'd need to better work on the naming, as it might be confusing.

So far they have different purposes; the `add` has just the purpose of keeping track of new spawned threads (basically fresh ones). Meanwhile `onWorkerAvailable` mostly scopes the fact that a Worker has just finished executing a task and might be candidate for a new task; so piscina can continue rolling up more tasks.

wdyt?

README.md
metcoder95: It's kind of hard to leverage this to the Scheduler, as Piscina does pretty much several abstractions for the threads when spawned.
Overall, the Scheduler can indicate to Piscina to spawn a new thread by returning null when calling `Scheduler#pick`.

README.md
metcoder95: It's to iterate over the threads spawned. I can actually thin this and replace it with an explicit API, I'll do that instead; anyways this is used only in a single place 👍

README.md
ronag: I think there might be an unnecessary overlap here. Does the scheduler even need to know when a thread is spawned? I think it only needs to know when a thread is available, period. No?

ronag: i.e. I think you can replace add/delete with workerAvailable/WorkerUnavailable or something.

ronag: another option is that you skip all of those, and pick receives an array of all available workers.

metcoder95:
>Any reason for going away from event emitters?

Do you mean for the `Scheduler` class?
If so, none in specific. Piscina already had, somehow, predefined this abstraction; so going in this way required fewer changes and breaking parts for Piscina. It also helped to standardise some pieces that were spread all around.

But open to suggestions, what do you have in mind?

README.md
metcoder95: >Does the scheduler even need to know when a thread is spawned? I think it only needs to know when a thread is available, period. No?

Hmm... see your point, but also the value of the Scheduler knowing when a thread is spawned. The idea is to allow the scheduler to have control to manage the threads at all times, so it knows its metrics, usage, and others and can decide which worker/thread to pick next of all available.
Piscina just takes control of spawning and counting that threads are within the boundaries of the options.

i.e. an example of a simple workload
maxThreads:2
minThreads:2

```txt
Piscina -> threads <= minThreads -> spawn new thread -> notify scheduler of new thread -> being a new`
```

maxThreads:3
minThreads:2
New incoming task
```txt
new task -> threads < maxThreads [yes] -> spawn new worker -> enqueue task
new task -> threads < maxThreads [no] -> Scheduler#pick -> use worker
```

Task Finished
```txt
worker -> piscina [task done] -> scheduler#onWorkerAvailable -> calls cbs to notify piscina a worker is available for new task -> piscina picks task from queue -> loops over new task workflow
```

But yeah, the naming is bad on the APIs, I'll change them 🤔

src/common.ts
metcoder95: Hmm, found the issue here; yeah, we just throw.
Before, as this was not abstracted, we just directly ignored the worker, but now we will need to notify the Scheduler about it so it can react accordingly.

metcoder95:
Thanks for the feedback @ronag, I'll work on those topics 👍

README.md
ronag: > so it knows its metrics, usage, and others and can decide which worker/thread to pick next of all available.

What's stopping it from starting to do from when it first becomes avaiable? Just keep a WeakMap<Worker, stats>

metcoder95: Yeah, nothing actually 🤔 

But then, do you suggest that is better to just make the Scheduler act as some sort of semaphore that hints Piscina what worker to use next?
And keep all the worker_thread management internally to piscina?

ronag: That's what I'm think based on this API... the current iteration is a bit in the middle gray zone and is a little ambigious where it's responsibility starts and ends....

ronag: The problem I guess is whether/how the scheduler tells piscina to create new threads or queue tasks.

metcoder95: Yeah, I get what you mean.
I've been considering offloading the thread instantiation to the Scheduler, but has as potential harm of killing all Piscina's purposes.

How do you imagine a way to tell Piscina to spawn a new thread or enqueue the task?
Currently is based on `Scheduler#pick` returning `null`; but stills Piscina will spawn more if the `maxThreads` hasn't been reached

ronag:
I was thinking a scheduler could look somehting like this:

```js
function makeBreathFirstScheduler(maxBreath) {
  const getCount = (worker) => worker.size
  return (task, workers) => {
    return workers.find(x => getCount(x) === 0)
      ?? (workers.length < maxBreath ? null : fp.sortBy(getCount, workers).at(0))
  }
}

function makeDepthFirstScheduler(maxDepth) {
  const getCount = (worker) => worker.size
  return (task, workers) => {
    return fp.sortBy(getCount, workers).findLast(x => getCount(x) < maxDepth)
  }
}

function makeLagScheduler(maxLag) {
  const getLag => (worker) => worker.lag
  return (task, workers) => {
    return fp.sortBy(getLag, workers).find(x => getLag(x) < maxLag)
  }
}
```

README.md
ronag: you could make pick or whatever return `Worker|SPAWN|QUEUE`

ronag: Yea, this abstraction is tricky

ronag: Maybe move the queue logic into the scheduler? So that it always either has to accept the task or ask for a new worker?

ronag: Alternatively we take the `maxThreads` setting?

metcoder95: >Yea, this abstraction is tricky

Yeah, trying to find the mid term is tricky 😅 

>Alternatively we take the maxThreads setting?

At the scheduler? Noup, it doesn't. But its something easy to add

>Maybe move the queue logic into the scheduler? So that it always either has to accept the task or ask for a new worker?

So, instead of `pick`, goes more into `Scheduler#run` and the scheduler decides to post the task?
That can be done, but pretty much a good portion of the magic of Piscina will be offloaded to the Scheduler, to coordinate the shared state, and prepare the task.
Most of the core of Piscina focuses on the coordination between thread and pool.

I think the Scheduler should be a thin layer that is aware of the threads, and balances the load between tasks.

>Worker|SPAWN|QUEUE

This seems like a good idea, to keep it simple, we can go ease as `0 = spawn`, `1 = queue`, `Worker = run task` 🤔

src/common.ts
metcoder95: Believe it or not, it seems to matter 😅 
If trying to offload it to the micro task or before the next tick, it hangs; it seems we just add stacks to the same tick and promises don't allow us to move forward in the loop; causing Piscina to hang.

Interesting gotcha

metcoder95:
After evaluating some tests issues, I noticed that although we do not modify the APIs, we alter the behaviour on startup of the Pool, which breaks the behavioral expectation, leading to a major bump.

I'm cleaning up the `next` branch as well as this one to point the changes to that branch instead

metcoder95:
After thinking about it more carefully, I believe I'll take a more conservative approach on this, noticed that overall the attempt here kind of breaks the nature of Piscina handling all the Worker orchestration beyond only customizing the load balancing part of it; I'll close the PR and look the problem from another angle.

I might reuse some of this, will hope to have something soon

Issue #497, opened 2024-01-28
Make Max Threads default to `os.availableParallelism()` https://nodejs.org/api/os.html#osavailableparallelism


metcoder95:
I think we could default it to `os.availableParallelism()` https://nodejs.org/api/os.html#osavailableparallelism

_Originally posted by @RafaelGSS in https://github.com/piscinajs/piscina/pull/487#discussion_r1467637405_

BenzeneAlcohol:
Hi, can I take this issue up?

metcoder95:
Please! 👍

PR #507, opened 2024-02-06
Added default value to maximumUsage in Scheduler

BenzeneAlcohol:
Added default value to maximumUsage

Fixes: https://github.com/piscinajs/piscina/issues/497

Something to do:

What if `maximumUsage > os.availableParallelism()` ?

### DIFF
--- a/src/scheduler.ts
+++ b/src/scheduler.ts
@@ -1,4 +1,5 @@
 import type { PiscinaWorker, PiscinaTask, RunTaskOptions } from '.';
+import * as os from 'os';
 
 function isTaskSchedulerLike (obj: {}): obj is TaskScheduler {
   if (Object.getPrototypeOf(obj) === TaskScheduler) return true;
@@ -66,10 +67,14 @@ class DefaultTaskScheduler extends TaskScheduler {
   #maximumUsage: number;
   #onAvailableListeners: ((item: PiscinaWorker) => void)[];
 
-  constructor (maximumUsage: number) {
+  constructor (maximumUsage: number = os.availableParallelism()) {
     super(maximumUsage);
     this.#maximumUsage = maximumUsage;
     this.#onAvailableListeners = [];
+
+    if (maximumUsage > os.availableParallelism()) {
+      console.warn(`Warning: maximumUsage (${maximumUsage}) is greater than available CPU cores (${os.availableParallelism()}).`);
+    }
   }
 
   add (item: PiscinaWorker) {


BenzeneAlcohol:
Also, `os.availableParallelism()` is available only on node versions > v18.14.0

src/scheduler.ts
```diff
@@ -66,10 +67,14 @@ class DefaultTaskScheduler extends TaskScheduler {
   #maximumUsage: number;
   #onAvailableListeners: ((item: PiscinaWorker) => void)[];
 
-  constructor (maximumUsage: number) {
+  constructor (maximumUsage: number = os.availableParallelism()) {
```
metcoder95, line 70: Although we do not have the matrix anymore, we still _officially_ support v16; so feel free to do a check over the Node version running and only get the value if the `os.availableParallelism` is available.

src/scheduler.ts
```diff
@@ -66,10 +67,14 @@ class DefaultTaskScheduler extends TaskScheduler {
...
     super(maximumUsage);
     this.#maximumUsage = maximumUsage;
     this.#onAvailableListeners = [];
+
+    if (maximumUsage > os.availableParallelism()) {
+      console.warn(`Warning: maximumUsage (${maximumUsage}) is greater than available CPU cores (${os.availableParallelism()}).`);
```
metcoder95, line 76: Feel free to use https://www.npmjs.com/package/process-warning instead, has better ergonomics for these kind of situations

metcoder95:
Let's also add a note in the documentation about this 👍

src/scheduler.ts
BenzeneAlcohol: So in case the version is less than 16, what should this default to? There is `os.cpus()` that almost does the same, do we default to the length of that, or some other number?

src/scheduler.ts
BenzeneAlcohol: Sure.

metcoder95:
Oh, my bad, I didn't notice you were working on top of my PR; let's aim first for the `current` version and later we can port it to the PR 👍 

So, for the current version, we should seek to use it whenever we want to get the max amount of threads if this option has not been passed through the constructor; so we can change it here: https://github.com/piscinajs/piscina/blob/a7042ea27d129f3cad75c422f5aa92f0663854ee/src/index.ts#L33

If the function is not available because it's being run under `Node v16`, let's keep the `cpus.length`.

The `maximumUsage`, is calculated based on the `concurrentTasksPerWorker`, which should be kept as default to the same number as `os.availableParallelism()` or `cpu.length`.

The warning, let's keep it but just port it to how is used within the current version of Piscina 👍

BenzeneAlcohol:
> Oh, my bad, I didn't notice you were working on top of my PR; let's aim first for the `current` version and later we can port it to the PR 👍
> 
> So, for the current version, we should seek to use it whenever we want to get the max amount of threads if this option has not been passed through the constructor; so we can change it here:
> 
> https://github.com/piscinajs/piscina/blob/a7042ea27d129f3cad75c422f5aa92f0663854ee/src/index.ts#L33
> 
> If the function is not available because it's being run under `Node v16`, let's keep the `cpus.length`.
> 
> The `maximumUsage`, is calculated based on the `concurrentTasksPerWorker`, which should be kept as default to the same number as `os.availableParallelism()` or `cpu.length`.
> 
> The warning, let's keep it but just port it to how is used within the current version of Piscina 👍

Right, got it!

PR #517, opened 2024-02-20
feat: allow named import usage

clydin:
The main exports of the package are now available both as named exports from the package as well as the previously available properties from the default export. This removes the requirement to use the `esModuleInterop` TypeScript option but retains backward compatiblity for existing usage.


src/entry.ts
metcoder95, line 1: Small `nit`. Can we rename it to `main.ts` that will be easier to grasp as the entry point of the library.

metcoder95 (dismissed):
Small `nit`, thanks for the contribution!

src/entry.ts
clydin: Updated.

metcoder95 (approved):

PR #518, opened 2024-02-20
feat: add option to disable run/wait time recording

clydin:
An additional `Piscina` constructor option is now available named `recordTiming` which can be used to control the underlying recording of wait and run times via Node.js histograms. By default this option is `true` to retain existing behavior but can be set to `false` to disable recording. The `Piscina` instance properties `runTime`, `waitTime`, and `utilization` will not be available if this option is disabled.
For longer lived processes or usage that does not rely on the time information, this removes the continuous storage of the information from memory.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -177,7 +177,8 @@ interface Options {
   taskQueue? : TaskQueue,
   niceIncrement? : number,
   trackUnmanagedFds? : boolean,
-  closeTimeout?: number
+  closeTimeout?: number,
+  recordTiming?: boolean
 }
 
 interface FilledOptions extends Options {
@@ -191,7 +192,8 @@ interface FilledOptions extends Options {
   useAtomics: boolean,
   taskQueue : TaskQueue,
   niceIncrement : number,
-  closeTimeout : number
+  closeTimeout : number,
+  recordTiming : boolean
 }
 
 const kDefaultOptions : FilledOptions = {
@@ -206,7 +208,8 @@ const kDefaultOptions : FilledOptions = {
   taskQueue: new ArrayTaskQueue(),
   niceIncrement: 0,
   trackUnmanagedFds: true,
-  closeTimeout: 30000
+  closeTimeout: 30000,
+  recordTiming: true
 };
 
 interface RunOptions {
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -206,7 +208,8 @@ const kDefaultOptions : FilledOptions = {
   taskQueue: new ArrayTaskQueue(),
   niceIncrement: 0,
   trackUnmanagedFds: true,
-  closeTimeout: 30000
+  closeTimeout: 30000,
+  recordTiming: true
```
metcoder95, line 212: I'll prepare a PR to set this to `false`, as soon for v5

src/index.ts
```diff
@@ -1267,14 +1276,26 @@ class Piscina extends EventEmitterAsyncResource {
   }
 
   get waitTime () : any {
+    if (!this.#pool.waitTime) {
+      throw Errors.TimeRecordingDisabled();
```
metcoder95: Let's not `throw` but return `null` instead.

In that way, we can set the signature of the getter to return either `HistogramSummary | null`

src/index.ts
```diff
@@ -1267,14 +1276,26 @@ class Piscina extends EventEmitterAsyncResource {
...
+    }
+
     return createHistogramSummary(this.#pool.waitTime);
   }
 
   get runTime () : any {
+    if (!this.#pool.runTime) {
```
metcoder95, line 1285: Same as my previous comment

src/index.ts
```diff
@@ -1267,14 +1276,26 @@ class Piscina extends EventEmitterAsyncResource {
...
+      throw Errors.TimeRecordingDisabled();
+    }
+
     return createHistogramSummary(this.#pool.runTime);
   }
 
   get utilization () : number {
+    if (!this.#pool.runTime) {
+      throw Errors.TimeRecordingDisabled();
```
metcoder95: ditto

src/index.ts
clydin: Since the type is already `any`, it could definitely return null instead of throwing.  Changing the type would be a breaking change though. All consumers, regardless of the new option's value, would need to add a guard before using the result with the more strict, although more accurate, return type.
Maybe the return type change would be a good v5 candidate as well?

metcoder95: Hmm, good point, that SGTM, let's keep it as is, and just return `null` instead, thanks for pointing that out!

metcoder95 (approved):

PR #554, opened 2024-05-02
docs: Replace third-party AbortController usage with native one in example

JaoodxD:
Given that the Readme specifies Piscina supports Node 16+, there is no longer a need to utilize a third-party `AbortController` implementation in the example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,6 @@ an `EventEmitter`:
 'use strict';
 
 const Piscina = require('piscina');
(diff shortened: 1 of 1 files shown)


JaoodxD:
@metcoder95 would you mind to review this little chore/improvement on docs?

metcoder95 (approved):

metcoder95:
Nice catch!

PR #555, opened 2024-05-04
feat: Use fixed queue

### LINKED ISSUE #452: Default TaskQueue implementation is slow for a large number of queued tasks

JaoodxD:
Some time ago @ronag mentioned in https://github.com/piscinajs/piscina/issues/452 that Piscina could adopt Nodejs fixed_queue as TaskQueue.

This draft will fix #452.

Synthetic benchmark shows drastical speed difference when comparing fixed_queue vs queue based on plain js array.
But actual [simple-benchmark.js](https://github.com/piscinajs/piscina/blob/current/benchmark/simple-benchmark.js) shows the oposite result.
I passed FixedQueue as custom queue [here](https://github.com/JaoodxD/piscina/blob/use-fixed-queue/benchmark/simple-benchmark-2.js).
Inital `simple-benchmark.js` shows around ~82k ops on my machine, but `simple-benchmark-2.js` with `FixedQueue` shows half of original performance with ~44k ops. 

Do I miss something?

### DIFF
--- a/README.md
+++ b/README.md
@@ -751,7 +751,7 @@ for (let i = 0; i < 10; i++) {
   });
 }
 ```
-In this example, we import `FixedQueue` from the `piscina` package and use it as the `taskQueue` option when creating a new Piscina pool. This allows users to take advantage of the performance improvements offered by `FixedQueue` before it becomes the default task queue implementation.
+**Note** The `FixedQueue` will become the default task queue implementation in a next major version.
 
 ## Current Limitations (Things we're working on / would love help with)
 
--- a/README.md
+++ b/README.md
@@ -728,7 +728,7 @@ In terms of Piscina performance itself, using `FixedQueue` with a queue size of 
 Users can import `FixedQueue` from the `Piscina` package and pass it as the `taskQueue` option to leverage its benefits.
 
 
-### Using FixedQueue Example
+#### Using FixedQueue Example
 
 Here's an example of how to use the `FixedQueue`:
 
--- a/README.md
+++ b/README.md
@@ -702,6 +702,56 @@ options on to the custom `TaskQueue` implementation. (Note that because the
 queue options are set as a property on the task, tasks with queue
 options cannot be submitted as JavaScript primitives).
 
+**New:** As of the next major release, Piscina introduces the `FixedQueue`, a more performant task queue implementation based on [`FixedQueue`](https://github.com/nodejs/node/blob/de7b37880f5a541d5f874c1c2362a65a4be76cd0/lib/internal/fixed_queue.js) from Node.Js project.  
+
+Here are some benchmarks to compare new `FixedQueue` with `ArrayTaskQueue` (current default). The benchmarks demonstrate substantial improvements in push and shift operations, especially with larger queue sizes.
+```
+Queue size = 1000
+┌─────────┬─────────────────────────────────────────┬───────────┬────────────────────┬──────────┬─────────┐
+│ (index) │ Task Name                               │ ops/sec   │ Average Time (ns)  │ Margin   │ Samples │
+├─────────┼─────────────────────────────────────────┼───────────┼────────────────────┼──────────┼─────────┤
+│ 0       │ 'ArrayTaskQueue full push + full shift' │ '9 692'   │ 103175.15463917515 │ '±0.80%' │ 970     │
+│ 1       │ 'FixedQueue  full push + full shift'    │ '131 879' │ 7582.696390658352  │ '±1.81%' │ 13188   │
+└─────────┴─────────────────────────────────────────┴───────────┴────────────────────┴──────────┴─────────┘
+
+Queue size = 100_000
+┌─────────┬─────────────────────────────────────────┬─────────┬────────────────────┬──────────┬─────────┐
+│ (index) │ Task Name                               │ ops/sec │ Average Time (ns)  │ Margin   │ Samples │
+├─────────┼─────────────────────────────────────────┼─────────┼────────────────────┼──────────┼─────────┤
+│ 0       │ 'ArrayTaskQueue full push + full shift' │ '0'     │ 1162376920.0000002 │ '±1.77%' │ 10      │
+│ 1       │ 'FixedQueue full push + full shift'     │ '1 026' │ 974328.1553396407  │ '±2.51%' │ 103     │
+└─────────┴─────────────────────────────────────────┴─────────┴────────────────────┴──────────┴─────────┘
+```
+In terms of Piscina performance itself, using `FixedQueue` with a queue size of 100,000 queued tasks can result in up to 6 times faster execution times.
+
+Users can import `FixedQueue` from the `Piscina` package and pass it as the `taskQueue` option to leverage its benefits.
+
+
+### Using FixedQueue Example
+
+Here's an example of how to use the `FixedQueue`:
+
+```js
+const { Piscina, FixedQueue } = require('piscina');
+const { resolve } = require('path');
+
+// Create a Piscina pool with FixedQueue
+const piscina = new Piscina({
+  filename: resolve(__dirname, 'worker.js'),
+  taskQueue: new FixedQueue()
+});
+
+// Submit tasks to the pool
+for (let i = 0; i < 10; i++) {
+  piscina.runTask({ data: i }).then((result) => {
+    console.log(result);
+  }).catch((error) => {
+    console.error(error);
+  });
+}
+```
+In this example, we import `FixedQueue` from the `piscina` package and use it as the `taskQueue` option when creating a new Piscina pool. This allows users to take advantage of the performance improvements offered by `FixedQueue` before it becomes the default task queue implementation.
+
 ## Current Limitations (Things we're working on / would love help with)
 
 * Improved Documentation
--- a/README.md
+++ b/README.md
@@ -702,7 +702,8 @@ options on to the custom `TaskQueue` implementation. (Note that because the
 queue options are set as a property on the task, tasks with queue
 options cannot be submitted as JavaScript primitives).
 
-**New:** As of the next major release, Piscina introduces the `FixedQueue`, a more performant task queue implementation based on [`FixedQueue`](https://github.com/nodejs/node/blob/de7b37880f5a541d5f874c1c2362a65a4be76cd0/lib/internal/fixed_queue.js) from Node.Js project.  
+### Built-In Queues
+Piscina also provides the `FixedQueue`, a more performant task queue implementation based on [`FixedQueue`](https://github.com/nodejs/node/blob/de7b37880f5a541d5f874c1c2362a65a4be76cd0/lib/internal/fixed_queue.js) from Node.js project.  
 
 Here are some benchmarks to compare new `FixedQueue` with `ArrayTaskQueue` (current default). The benchmarks demonstrate substantial improvements in push and shift operations, especially with larger queue sizes.
 ```
--- a/package.json
+++ b/package.json
@@ -16,11 +16,13 @@
     "test:ci": "npm run lint && npm run build && npm run test:coverage",
     "test:coverage": "c8 --reporter=lcov tap --cov",
     "prepack": "npm run build",
-    "bench": "npm run benchmark:piscina-default && npm run benchmark:piscina-fixed-queue && npm run benchmark:queue-comparison && npm run benchmark:piscina-comparison",
+    "bench": "npm run bench:taskqueue && npm run bench:piscina",
+    "bench:piscina": "npm run benchmark:piscina-default &&npm run benchmark:piscina-fixed-queue && npm run benchmark:piscina-comparison",
+    "bench:taskqueue": "npm run benchmark:queue-comparison",
     "benchmark:piscina-default": "node benchmark/simple-benchmark.js",
     "benchmark:piscina-fixed-queue": "node benchmark/simple-benchmark-fixed-queue.js",
-    "benchmark:queue-comparison": "node benchmark/queue-comparison.js",
-    "benchmark:piscina-comparison": "node benchmark/piscina-queue-comparison.js"
+    "benchmark:piscina-comparison": "node benchmark/piscina-queue-comparison.js",
+    "benchmark:queue-comparison": "node benchmark/queue-comparison.js"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -15,7 +15,12 @@
     "test": "c8 tap",
     "test:ci": "npm run lint && npm run build && npm run test:coverage",
     "test:coverage": "c8 --reporter=lcov tap --cov",
-    "prepack": "npm run build"
+    "prepack": "npm run build",
+    "bench": "npm run benchmark:piscina-default && npm run benchmark:piscina-fixed-queue && npm run benchmark:queue-comparison && npm run benchmark:piscina-comparison",
+    "benchmark:piscina-default": "node benchmark/simple-benchmark.js",
+    "benchmark:piscina-fixed-queue": "node benchmark/simple-benchmark-fixed-queue.js",
+    "benchmark:queue-comparison": "node benchmark/queue-comparison.js",
+    "benchmark:piscina-comparison": "node benchmark/piscina-queue-comparison.js"
   },
   "repository": {
     "type": "git",
@@ -44,6 +49,7 @@
     "snazzy": "^9.0.0",
     "standardx": "^7.0.0",
     "tap": "^16.3.7",
+    "tinybench": "^2.8.0",
     "ts-node": "^10.9.2",
     "typescript": "5.4.5"
   },
--- a/package.json
+++ b/package.json
@@ -15,7 +15,12 @@
     "test": "tap --ts",
     "test:ci": "npm run lint && npm run build && npm run test:coverage",
     "test:coverage": "tap --ts --cov --coverage-report=html --no-browser --no-check-coverage",
-    "prepack": "npm run build"
+    "prepack": "npm run build",
+    "bench": "npm run benchmark:piscina-default && npm run benchmark:piscina-fixed-queue && npm run benchmark:queue-comparison && npm run benchmark:piscina-comparison",
+    "benchmark:piscina-default": "node benchmark/simple-benchmark.js",
+    "benchmark:piscina-fixed-queue": "node benchmark/simple-benchmark-fixed-queue.js",
+    "benchmark:queue-comparison": "node benchmark/queue-comparison.js",
+    "benchmark:piscina-comparison": "node benchmark/piscina-queue-comparison.js"
   },
   "repository": {
     "type": "git",
@@ -43,10 +48,10 @@
     "snazzy": "^9.0.0",
     "standardx": "^7.0.0",
     "tap": "^16.3.7",
+    "tinybench": "^2.8.0",
     "ts-node": "^10.9.2",
     "typescript": "5.4.5"
   },
-  "dependencies": {},
   "optionalDependencies": {
     "nice-napi": "^1.0.2"
   },
--- a/src/fixed-queue.ts
+++ b/src/fixed-queue.ts
@@ -153,7 +153,7 @@ export default class FixedQueue implements TaskQueue {
   }
 
   remove (task: Task) {
-    let prev = null;
+    let prev: FixedCircularBuffer | null = null;
     let buffer = this.tail;
     while (true) {
       if (buffer.list.includes(task)) {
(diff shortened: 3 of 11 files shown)


JaoodxD:
@ronag @metcoder95 could you take a look?

src/fixed-queue.ts
metcoder95, line 1: Can you add a reference to the original code at Node's codebase, as well as its license?

metcoder95 (commented):
It can potentitally due to the optimizations done for Array operations. Haven't check but often they are already inlined within v8 (sometimes even in assembly already).

Can you try to use a tool like `tinybench` to benchmark them?

src/fixed-queue.ts
JaoodxD: Sure

JaoodxD: Thanks to dexnode and Deopt Explorer extension I finnaly starting to realize what I've done wrong. 
First problem was `size` getter of queue, it was accessed massively so I rewrote
```ts
get size () {
     let count = 0;
     for (let i = 0; i < this.list.length; i++) {
       if (this.list[i] !== undefined) count++;
     }
     return count;
  }
```
Now the performance becomes equal (~82k opts vs ~82k opts).
Hope I'll find all the bottlenecks.

JaoodxD: > Can you add a reference to the original code at Node's codebase, as well as its license?

Please check if committed mention of original nodejs implementation is sufficient

metcoder95 (commented):
Let's export the new `FixedQueue` as a built-in task queue that can be used instead of the current one, so users have the chance to use it.

Let's add some documentation for it as well, and it should be good to go

ronag:
We could also make a "smart" mode. Where if more than x (128?) number of items have been queued we switch automatically to fixed queue.

metcoder95:
Nice, that would be great! We can add an option called `smart` being boolean, `false` (which is also the default) keeps using the same Task queue and `true`, makes the switch to `FixedQueue` after a threshold (128 seems like a good default to start with).

Tho if the default task queue is already a FixedQueue, there's nothing to do there.

JaoodxD:
Although the bottleneck with size getter is already fixed, I still can't find the main one. 
Tinybench benchmarks shows strange behavior.
When I compare just two queues, FixedQueue behaves faster, in order of magnitudes faster and the order rises with amount of elements in it.
But when I do the same benchmarks against Piscina instances, it shows identical perf, like queue doesn't matter at all.

ronag:
TBH. I think we should always use FixedQueue by default, the "smart" mode is just a compromise suggestion

JaoodxD:
> Although the bottleneck with size getter is already fixed, I still can't find the main one. Tinybench benchmarks shows strange behavior. When I compare just two queues, FixedQueue behaves faster, in order of magnitudes faster and the order rises with amount of elements in it. But when I do the same benchmarks against Piscina instances, it shows identical perf, like queue doesn't matter at all.

Here's a benchmark of 1000 items pushed to queue and then shifted:
```
┌─────────┬─────────────────────────────────────────┬───────────┬────────────────────┬──────────┬─────────┐
│ (index) │ Task Name                               │ ops/sec   │ Average Time (ns)  │ Margin   │ Samples │
├─────────┼─────────────────────────────────────────┼───────────┼────────────────────┼──────────┼─────────┤
│ 0       │ 'ArrayTaskQueue full push + full shift' │ '9 692'   │ 103175.15463917515 │ '±0.80%' │ 970     │
│ 1       │ 'FixedQueue  full push + full shift'    │ '131 879' │ 7582.696390658352  │ '±1.81%' │ 13188   │
└─────────┴─────────────────────────────────────────┴───────────┴────────────────────┴──────────┴─────────┘
```
We can see that FixedQueue is ~13 times faster.

But when we do the same benchmark against Piscina with two different `taskQueue` it shows following:
```
┌─────────┬─────────────────────────────────────────┬─────────┬────────────────────┬──────────┬─────────┐
│ (index) │ Task Name                               │ ops/sec │ Average Time (ns)  │ Margin   │ Samples │
├─────────┼─────────────────────────────────────────┼─────────┼────────────────────┼──────────┼─────────┤
│ 0       │ 'ArrayTaskQueue full push + full shift' │ '7'     │ 132713349.99999994 │ '±2.11%' │ 10      │
│ 1       │ 'FixedQueue  full push + full shift'    │ '7'     │ 129970319.99999999 │ '±1.93%' │ 10      │
└─────────┴─────────────────────────────────────────┴─────────┴────────────────────┴──────────┴─────────┘
```
Which is very strange.

JaoodxD:
Should I push this benchmarks into the branch?

ronag:
I think the difference is within the margin of error.

JaoodxD:
> I think the difference is within the margin of error.

And I completely don't understand why, I thought that even at this scale the difference should be measurable.

I just raised the amount of tasks to be queued to 100_000 and Piscina benchmark started to show the difference.
```
┌─────────┬─────────────────────────────────────────┬─────────┬────────────────────┬──────────┬─────────┐
│ (index) │ Task Name                               │ ops/sec │ Average Time (ns)  │ Margin   │ Samples │
├─────────┼─────────────────────────────────────────┼─────────┼────────────────────┼──────────┼─────────┤
│ 0       │ 'ArrayTaskQueue full push + full shift' │ '0'     │ 8022717600.000002  │ '±4.37%' │ 10      │
│ 1       │ 'FixedQueue  full push + full shift'    │ '0'     │ 1304727259.9999983 │ '±4.89%' │ 10      │
└─────────┴─────────────────────────────────────────┴─────────┴────────────────────┴──────────┴─────────┘
```
FixedQueue is ~6 times faster now.

JaoodxD:
Seems like I have found another caveat. Typescript target ES2019 compiles private fields into something that kills performance.

TLDR: 
ES2022 private fields x6 times faster, than ES2019 (Current project target). 
Old-style private fields with `_` is x7 faster on the same ES2019 target, and ~20% faster than private fields on ES2022 target.

Here is benchmark comparison of current `ArrayTaskQueue` vs `FixedQueue` with target: ES2019
```
┌─────────┬─────────────────────────────────────────┬─────────┬────────────────────┬──────────┬─────────┐
│ (index) │ Task Name                               │ ops/sec │ Average Time (ns)  │ Margin   │ Samples │
├─────────┼─────────────────────────────────────────┼─────────┼────────────────────┼──────────┼─────────┤
│ 0       │ 'ArrayTaskQueue full push + full shift' │ '0'     │ 1143968079.9999998 │ '±0.48%' │ 10      │
│ 1       │ 'FixedQueue full push + full shift'     │ '146'   │ 6835453.33333362   │ '±3.78%' │ 15      │
└─────────┴─────────────────────────────────────────┴─────────┴────────────────────┴──────────┴─────────┘
```
And here's the same benchmark with target ES2022:
```
┌─────────┬─────────────────────────────────────────┬─────────┬────────────────────┬──────────┬─────────┐
│ (index) │ Task Name                               │ ops/sec │ Average Time (ns)  │ Margin   │ Samples │
├─────────┼─────────────────────────────────────────┼─────────┼────────────────────┼──────────┼─────────┤
│ 0       │ 'ArrayTaskQueue full push + full shift' │ '0'     │ 1139727079.9999993 │ '±0.42%' │ 10      │
│ 1       │ 'FixedQueue full push + full shift'     │ '849'   │ 1176970.930232507  │ '±2.56%' │ 86      │
└─────────┴─────────────────────────────────────────┴─────────┴────────────────────┴──────────┴─────────┘
```
And that's what we can get with target ES2019 but rewriting modern private field syntax into old one with underscore (like `_field`):
```
┌─────────┬─────────────────────────────────────────┬─────────┬────────────────────┬──────────┬─────────┐
│ (index) │ Task Name                               │ ops/sec │ Average Time (ns)  │ Margin   │ Samples │
├─────────┼─────────────────────────────────────────┼─────────┼────────────────────┼──────────┼─────────┤
│ 0       │ 'ArrayTaskQueue full push + full shift' │ '0'     │ 1162376920.0000002 │ '±1.77%' │ 10      │
│ 1       │ 'FixedQueue full push + full shift'     │ '1 026' │ 974328.1553396407  │ '±2.51%' │ 103     │
└─────────┴─────────────────────────────────────────┴─────────┴────────────────────┴──────────┴─────────┘
```
As we can see the same FixedQueue implementation shows different results.
* 146 ops/sec with `"target" : "ES2019"` and modern private fields syntax.
* 849 ops/sec with `"target" : "ES2022"` and modern private fields syntax.
* 1026 ops/sec with `"target" : "ES2019"` and old-style agreement of using _ in the name of "private" properties.

And the question is should we update target or rewrite all private fields on the hotpath?

metcoder95:
> TBH. I think we should always use FixedQueue by default, the "smart" mode is just a compromise suggestion

I'm afraid of breaking implementation if there is custom usage of the current ArrayQueue, I'm ok with keeping it like this within the current version, and make it the default on `major`.

I'm planning to expedite a major mid June or early July

>FixedQueue is ~6 times faster now.

It's kind of expected.
FixedQueue behaves better over a big load, compared to the ArrayQueue, that's why the suggestion of the mode is based on a threshold.

>And the question is should we update target or rewrite all private fields on the hotpath?

Let's rewrite it for now, and we can aim to target 2022 private syntax on next major

JaoodxD:
So I propose to do following:
1. Expose to import/require `FixedQueue` from the module.
2. Mention it in docs so the users are able to pass it as custom `taskQueue` as demonstrated below:
```js
import { Piscina, FixedQueue } from 'Piscina'

const pool = new Piscina({
  filename: resolve(__dirname, 'fixtures/add.js'),
  taskQueue: new FixedQueue()
})
```

> and make it the default on major.

Until it eventualy becomes default in the following major release.

JaoodxD:
Also I have a few little benchmarks (results of which I posted above) and `tinybench` as new dev dependency. 
Should I commit them or it is not necessary?

JaoodxD:
> Nice, that would be great! We can add an option called `smart` being boolean, `false` (which is also the default) keeps using the same Task queue and `true`, makes the switch to `FixedQueue` after a threshold (128 seems like a good default to start with).
> 
> Tho if the default task queue is already a FixedQueue, there's nothing to do there.

I also think that `FixedQueue` would be better in all cases.
The only benefit current ArrayQueue has is memory footprint, since FixedQueue preallocate 2048 elements array, while ArrayQueue allocate only 4 on first insertion.

metcoder95:
>So I propose to do following:

SGTM 👍 

>Should I commit them or it is not necessary?

Let's commit them, will serve as a baseline for further optimizations.

>I also think that FixedQueue would be better in all cases.
The only benefit current ArrayQueue has is memory footprint, since FixedQueue preallocate 2048 elements array, while ArrayQueue allocate only 4 on first insertion.

Agree, let's mention this while documenting the new `FixedQueue` 👍

JaoodxD:
I've faced an issue with named exports in commonjs target. 
It turns out that current version completely skips them due to following code in `src/main.ts`:
```ts
import Piscina from './index';

// Used as the require() entry point to maintain existing behavior
export = Piscina;
```
It leads to following behavior: typescript compiler properly generate types but actual exported fields and classes (named exports) are missing.

JaoodxD:
May we mixin them to `Piscina` class in `src/index.ts` like following?
```ts
...
// Mutate Piscina class to allow named import in commonjs
// @ts-expect-error
Piscina.FixedQueue = FixedQueue;
// @ts-expect-error
Piscina.ArrayTaskQueue = ArrayTaskQueue;
```

To allow 
```js
const { Piscina, FixedQueue } = require('piscina');
```

metcoder95:
Yeah, that's the way to go 👍

JaoodxD:
Find out some issues with `FixedQueue.remove` currently working on resolving it.

JaoodxD:
Finally I fixed it.

JaoodxD:
Also I find out that Node's implementation has a bug. 
Every Circular buffer of FixedQueue actually had 1 empty element.
Caused by [this line of code](https://github.com/nodejs/node/blob/9807ede6fb17afe36a2447df65eb6b63df8d1d37/lib/internal/fixed_queue.js#L71-L73):
```js
 isFull() {
    return ((this.top + 1) & kMask) === this.bottom;
  }
```

When we fulfill Circular buffer for 2047 of 2048 elements and trying to push another one, it will create another Circular buffer so we have first buffer with 2047 elements and 1 empty slot + new buffer with 1 element and 2047 empty slots. 

***
Looks like a potential PR to safe some memory for Node.js itself

metcoder95:
Can you rebase with `current`?

JaoodxD:
> Can you rebase with `current`?

Sure

metcoder95:
For the suggestion of the `FixedQueue` issue, I'd advice you to open an issue in Node for that; its possible it is by design, otherwise a issue to fix 👍

JaoodxD:
> For the suggestion of the `FixedQueue` issue, I'd advice you to open an issue in Node for that; its possible it is by design, otherwise a issue to fix 👍

I sat and thought for a while and came to the same conclusion. Maybe this is done in order not to check the real number of elements.
But it will be interesting to know for sure.

README.md
```diff
@@ -702,6 +702,56 @@ options on to the custom `TaskQueue` implementation. (Note that because the
 queue options are set as a property on the task, tasks with queue
 options cannot be submitted as JavaScript primitives).
 
+**New:** As of the next major release, Piscina introduces the `FixedQueue`, a more performant task queue implementation based on [`FixedQueue`](https://github.com/nodejs/node/blob/de7b37880f5a541d5f874c1c2362a65a4be76cd0/lib/internal/fixed_queue.js) from Node.Js project.
```
metcoder95:
```suggestion
### Built-In Queues
Piscina also provides the `FixedQueue`, a more performant task queue implementation based on [`FixedQueue`](https://github.com/nodejs/node/blob/de7b37880f5a541d5f874c1c2362a65a4be76cd0/lib/internal/fixed_queue.js) from Node.js project.  
```

README.md
```diff
@@ -702,6 +702,56 @@ options on to the custom `TaskQueue` implementation. (Note that because the
...
+│ 0       │ 'ArrayTaskQueue full push + full shift' │ '0'     │ 1162376920.0000002 │ '±1.77%' │ 10      │
+│ 1       │ 'FixedQueue full push + full shift'     │ '1 026' │ 974328.1553396407  │ '±2.51%' │ 103     │
+└─────────┴─────────────────────────────────────────┴─────────┴────────────────────┴──────────┴─────────┘
+```
+In terms of Piscina performance itself, using `FixedQueue` with a queue size of 100,000 queued tasks can result in up to 6 times faster execution times.
+
+Users can import `FixedQueue` from the `Piscina` package and pass it as the `taskQueue` option to leverage its benefits.
+
+
+### Using FixedQueue Example
```
metcoder95:
```suggestion
#### Using FixedQueue Example
```

README.md
```diff
@@ -702,6 +702,56 @@ options on to the custom `TaskQueue` implementation. (Note that because the
...
+// Submit tasks to the pool
+for (let i = 0; i < 10; i++) {
+  piscina.runTask({ data: i }).then((result) => {
+    console.log(result);
+  }).catch((error) => {
+    console.error(error);
+  });
+}
+```
+In this example, we import `FixedQueue` from the `piscina` package and use it as the `taskQueue` option when creating a new Piscina pool. This allows users to take advantage of the performance improvements offered by `FixedQueue` before it becomes the default task queue implementation.
```
metcoder95:
```suggestion
**Note** The `FixedQueue` will become the default task queue implementation in a next major version.
```

package.json
```diff
@@ -15,7 +15,12 @@
     "test": "c8 tap",
     "test:ci": "npm run lint && npm run build && npm run test:coverage",
     "test:coverage": "c8 --reporter=lcov tap --cov",
-    "prepack": "npm run build"
+    "prepack": "npm run build",
+    "bench": "npm run benchmark:piscina-default && npm run benchmark:piscina-fixed-queue && npm run benchmark:queue-comparison && npm run benchmark:piscina-comparison",
+    "benchmark:piscina-default": "node benchmark/simple-benchmark.js",
+    "benchmark:piscina-fixed-queue": "node benchmark/simple-benchmark-fixed-queue.js",
+    "benchmark:queue-comparison": "node benchmark/queue-comparison.js",
+    "benchmark:piscina-comparison": "node benchmark/piscina-queue-comparison.js"
```
metcoder95: Let's maybe split these and aggregate them into two different scopes `bench:taskqueue` and we attempt to run all the TaskQueue related benchmarks, and left `bench` to run all benchmarks.

We might add more with the work over the new load balancer and scheduler that I'd like to work on soon.

test/fixed-queue.ts
metcoder95, line 1: Can you add some tests using the task queue with `Piscina`?

package.json
JaoodxD: Good point

JaoodxD: Done

test/fixed-queue.ts
JaoodxD: Added a few. Should I test any particular cases?

JaoodxD:
> For the suggestion of the `FixedQueue` issue, I'd advice you to open an issue in Node for that; its possible it is by design, otherwise a issue to fix 👍 

Seems like it wasn't a bug and I should've carefully read the comments which clearly states the following:
```
// When `top === bottom` the current queue is empty and when
// `top + 1 === bottom` it's full. This wastes a single space of storage
// but allows much quicker checks.
```
https://github.com/nodejs/node/blob/9807ede6fb17afe36a2447df65eb6b63df8d1d37/lib/internal/fixed_queue.js#L55

***
Will double check how it affects performance and then revert my "fixes" if the difference is observable.

test/fixed-queue.ts
metcoder95: Let's go with basics and possible cases that you'd like to test when you implemented the new task queue

metcoder95:
>Will double check how it affects performance and then revert my "fixes" if the difference is observable.

Give it a try with max size + 1, and (max * 2) + 1 as baseline just to see its impact when expanding the queue

metcoder95 (approved):

JaoodxD:
@metcoder95 sorry. Been busy with work, so haven't added any additional benchmarks.
Will make a separate PR with them a bit later.

metcoder95:
np! The PR was great to land already, as it is not the default the benchmarks are great in separated PR 👍

JaoodxD:
@metcoder95 Not quietly sure where and in which shape it should be present in piscina repo (if should at all?).
Here's a result of benchmark from following repo.  
https://github.com/JaoodxD/fixed-queue-perf-test  
***
It shows results for queue size of
- max_size + 1 (Current Piscina FixedQueue is 15% faster)
- max_size * 2 + 1 (Node is faster for 20%)
- max_size * 64 + 1 (Node is faster for 13%)
- max_size * 1024 + 1 (Node is faster for 18%)  

As we might notice the first suite shows that Piscina fix actualy has positive effect, but it's an edge case and in general Node's implementation is roughly 13-20% faster.

image

metcoder95:
@JaoodxD would you like to continue the research about the private properties?

JaoodxD:
> @JaoodxD would you like to continue the research about the private properties?

I would like to.

JaoodxD:
@metcoder95 quick test showed a good news and a bad news. 
A good news is the fact that both `_field` and `#field` have identical performance with `ES2022` and newer targets. 
The bad one is we get 17-20% perf regression in comparison to what we have right now.

JaoodxD:
> @metcoder95 quick test showed a good news and a bad news. A good news is the fact that both `_field` and `#field` have identical performance with `ES2022` and newer targets. The bad one is we get 17-20% perf regression in comparison to what we have right now.

And I found a way to fix it

metcoder95:
Feel free to open a PR, and also is safe to assume we will target ES2022 at minimum in v5

PR #558, opened 2024-05-05
chore: split structure

metcoder95:
- **chore: use node-prefix**
- **chore: consolidate common**
- **chore: scope implementations together**

Attempt to try to reduce the complexity of the project with the goal of ease of understanding

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,12 +1,14 @@
-import { Worker, MessageChannel, MessagePort, receiveMessageOnPort } from 'worker_threads';
-import { once, EventEmitterAsyncResource } from 'events';
-import { AsyncResource } from 'async_hooks';
-import { availableParallelism } from 'os';
-import { fileURLToPath, URL } from 'url';
-import { resolve } from 'path';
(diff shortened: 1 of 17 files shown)


metcoder95:
I'll be merging, please @RafaelGSS if you have any feedback I'll address it in a further PR 👍

src/index.ts
```diff
@@ -1,158 +1,67 @@
...
-  get size () { return this.tasks.length; }
-
-  shift () : Task | null {
-    return this.tasks.shift() as Task;
-  }
-
-  push (task : Task) : void {
-    this.tasks.push(task);
-  }
+const { version } = JSON.parse(
```
michaelangeloio, line 56: @metcoder95 hi, this is currently breaking things for us. 

Is this `version` extraction required? can we pass in a custom path or version in options instead?

This breaks flexibility in a number of ways, like using esbuild and bundling.

metcoder95: Let me revert, this is causing too many problems. Will open a PR and publish a fix soon

PR #565, opened 2024-05-15
feat: expose task interface

metcoder95:
- **chore: split worker pool and task queue implementation**
- **refactor: remove leftovers**


metcoder95:
I'm trying to decouple the Task Queue and Worker Pool implementation so they can grow independently, mostly aiming for enable the work for  #536 and do it progressively; I'd target the new balancer work directly to `next` as safe bet

RafaelGSS (commented):
Is it semver-major?

metcoder95:
Don't think so, wasn't exported before, but if we want to be careful, we can do so

metcoder95:
Just wanted to verify if is true that we did not expose these types, and indeed we don't.
The only form to have access to them was through

```ts
import type { Task } from 'piscina/dist/types';
```

Which is not officially supported, instead to access the same types they can do:
```ts
import type { PiscinaTask } from 'piscina';
```

metcoder95:
I'll be merging the PR, if there's extra feedback, feel free to add it here and I'll address it in further PR

PR #569, opened 2024-05-20
feat: allow generic when creating Piscina

### LINKED ISSUE #231: Use generic Piscina class to typecheck `run()` parameters and return type.
Changing the Piscina class to use a generic `T` (parameter) and `R` (possible return). This way it would be possible to get a better developer experience and intellisense while using this package.

Example: 

```ts
declare class Piscina<T = any, R = any> ... {
  ...

  run(task: T, options?: RunOptions): Promise<R>;
}
```

I'm also open to create a PR...

Thanks!

alexframe:
Add ability to pass `Task` and `Return` type when creating `Piscina` class.

Example usage:

```typescript
const pool = new Piscina<MyTask, MyResult>();

// result resolves as `MyResult` and the args is expected to be a `MyTask`
const result = await pool.run({ my: 'task' }); 
```

Resolves #231.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1075,7 +1075,7 @@ class ThreadPool {
   }
 }
 
-export default class Piscina extends EventEmitterAsyncResource {
+export default class Piscina<T = any, R = any> extends EventEmitterAsyncResource {
   #pool : ThreadPool;
 
   constructor (options : Options = {}) {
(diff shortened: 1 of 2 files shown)


metcoder95 (commented):
Can you add some testing to it?

alexframe:
@metcoder95 I have added a couple of tests, two of which are annotated with `@ts-expect-error` to ensure that an error is thrown by typescript when an incorrect type is assigned.

Let me know if there's anything else you think is needed.

metcoder95 (approved):
Thanks for the contribution! LGTM

PR #574, opened 2024-05-22
chore(perf): Revert commit to bring back up to 20% perf

JaoodxD:
As mentioned here https://github.com/piscinajs/piscina/pull/555#issuecomment-2125000944 in one of the commits capacity optimization were made.  
It saved 1 unit of storage per circular buffer but actually takes up to 20% of performance.  
Link to commit which caused perf drop: https://github.com/piscinajs/piscina/pull/555/commits/395e630e2f9791fdc04fbc05735694ff251c06a5

### DIFF
--- a/test/fixed-queue.ts
+++ b/test/fixed-queue.ts
@@ -63,8 +63,18 @@ test('remove not queued task should not lead to errors', async ({ equal }) => {
 });
 
 test('removing elements from intermediate CircularBuffer should not lead to issues', async ({ equal, same }) => {
+  /*
+      The test intends to check following scenario:
+      1) We fill the queue with 3 full circular buffers amount of items.
+      2) Empty the middle circular buffer with remove().
+      3) This should lead to the removal of the middle buffer from the queue:
+         - Before emptying: tail buffer -> middle buffer -> head buffer.
+         - After emptying: tail buffer -> head buffer.
+   */
+
   const queue = new FixedQueue();
 
+  // size of single circular buffer
   const batchSize = 2047;
 
   const firstBatch = Array.from({ length: batchSize }, () => new QueueTask());
@@ -94,8 +104,17 @@ test('removing elements from intermediate CircularBuffer should not lead to issu
 });
 
 test('removing elements from first CircularBuffer should not lead to issues', async ({ equal, same }) => {
+  /*
+      The test intends to check following scenario:
+      1) We fill the queue with 3 full circular buffers amount of items.
+      2) Empty the first circular buffer with remove().
+      3) This should lead to the removal of the tail buffer from the queue:
+         - Before emptying: tail buffer -> middle buffer -> head buffer.
+         - After emptying: tail buffer (previously middle) -> head buffer.
+   */
   const queue = new FixedQueue();
 
+  // size of single circular buffer
   const batchSize = 2047;
 
   const firstBatch = Array.from({ length: batchSize }, () => new QueueTask());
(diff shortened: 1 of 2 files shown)


test/fixed-queue.ts
metcoder95, line 1: Can we do benchmarks with payloads twice or three times its size?

JaoodxD: Do you mean to push something bigger than integers or just more items?
If the second option then it's already done. 
Results could be seen here https://github.com/piscinajs/piscina/pull/555#issuecomment-2125000944 for twice the size, 64 times the size and 1024 times the size of single circular buffer.

metcoder95: Oh, good call, sorry I missed that

JaoodxD: No problem.
Btw in the last few days I tried to play with circular buffer size, tried to preallocate and reuse ring buffers with a pool and so on but nothing really brings benefits, only performance degradations.  
It feels like FixedQueue was created by a dark magician who put an "unbeatable" spell on it. 😅

metcoder95: It is pretty well optimized. Liked the tradeoff of memory for more performance 👌

test/fixed-queue.ts
```diff
@@ -65,7 +65,7 @@ test('remove not queued task should not lead to errors', async ({ equal }) => {
 test('removing elements from intermediate CircularBuffer should not lead to issues', async ({ equal, same }) => {
   const queue = new FixedQueue();
 
-  const batchSize = 2048;
+  const batchSize = 2047;
```
metcoder95, line 78: Does it fails with 2048?
In the worst case, it will pre-allocate another buffer, isn't it?

JaoodxD: In general it doesn't fail but it should match the batch size due to test's intensions.  
The test may require additional clarification with comments.  
The main goal was to check following scenario:
1. We fill the queue with 3 full circular buffers amount of items.
2. Empty the middle circular buffer.
3. This should lead to the removal of the middle buffer from the queue:
  - Before emptying: tail buffer -> middle buffer -> head buffer.
  - After emptying: tail buffer -> head buffer.

My initial `remove` implementation didn't remove empty buffer so I wrote a test to fix it.

metcoder95: Nice, now is clear; maybe add that to the description of the test can help understand why are we doing that

JaoodxD: Although I added the description I must admit that the test is actually insufficient/fragile due to lack of internal checks.  
As you rightly noticed it doesn't and it wouldn't break even if the size of internal circular buffer will ever changes.  
We potentialy could try to really check the amount of buffers with something like following, but this approach will definetly be broken when we ship "real" private fields in further major release.  
```js
const queue = new FixedQueue()

let buffer = queue.tail
let count = 0

while (buffer !== null) {
  count++
  buffer = buffer.next
}

console.log(`The queue has ${count} buffers`)
```

metcoder95: Overall, I'd not recommend to alter the internals just for testing; I've always advocated for testing the outcome rather than expectations.
If we want to check how many internal circular buffers has been made, we can always check the overall outcome, we know that if two circular buffers are made, at least we should have 2048 + 1, this at the `FixedTaskQueue` level.

At the `FixedQueue` we can go deep but always using the interfaces as a bridge and not going beyond

JaoodxD: I completely agree that it's not the right way. I just worry about further consequences due to the lack of clarity. 
Is there anything else blocking the PR from being landed?

metcoder95 (approved):

PR #590, opened 2024-06-12
feat: Custom Balancer

### LINKED ISSUE #347: More customisable scheduling

metcoder95:
- [x] Add histograms to workers - #619 
- [x] Expose workers lifecycle events - #620 
- [x] Expose pool lifecycle events - #624
- [x] Decouple handling of when new worker available and when a worker has finished a task
- [x] ~~Abstract event handling through AsyncResourcePool~~
- [x] Decouple LB from scheduler (LB should return either `worker` or `null`)
- [x] Implement LB branching
	- [x] LB on new worker
	- [x] LB once worker finished task
	- [x] LB on new task


metcoder95:
Still need to figure out a set of things but it should be on a good shape to bring up the idea of how the balancer can look like. Let me know what you think, and I'll adjust over the week 👍

src/worker_pool/base.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  maybeAvailable (item : T) {
+    /* istanbul ignore else */
+    if (item.currentUsage() < this.maximumUsage) {
+      for (const listener of this.onAvailableListeners) {
+        listener(item);
+      }
+    }
+  }
```
ronag, line 100: What is the purpose of:
- iterator
- maybeAvailable

metcoder95: - `maybeAvailable` -- It is meant to hint that a worker is available, either because has been bootstrapped recently or has finished executing a task
- `iterator` -- to support the pool iterate over the workers, e.g. to process pending messages (on teardown)

src/index.ts
```diff
@@ -32,7 +33,10 @@ import {
 } from './task_queue';
 import {
   WorkerInfo,
-  AsynchronouslyCreatedResourcePool
+  AsynchronouslyCreatedResourcePool,
+  PisicnaLoadBalancer,
```
jerome-benoit:
```suggestion
  PiscinaLoadBalancer,
```

src/worker_pool/balancer/index.ts
```diff
@@ -0,0 +1,43 @@
+import type { PiscinaTask } from '../../task_queue';
+import type { PiscinaWorker } from '..';
+
+export type PiscinaLoadBalancerCommand = {
+  candidate?: PiscinaWorker | null; // If candidate is passed, it will be used as the result of the load balancer and ingore the command
+  command?: 0 | 1; // 1 = add, 0 = busy
+};
+export type PisicnaLoadBalancer = (
+  task: PiscinaTask,
+  workers: PiscinaWorker[]
```
jerome-benoit, line 6: Are the PiscinaWorker[] exposing their usage statistics? Without them, the load balancer code is not able to take any decision besides simple algorithms like round robin. And since the usage statistics are currently dispatched in different thread pool private properties, adding them to the API without refactoring will end with an quite ugly LB API.

src/worker_pool/index.ts
```diff
@@ -5,112 +5,19 @@ import { RequestMessage, ResponseMessage } from '../types';
...
+type ResponseCallback = (response : ResponseMessage) => void;
 
-    return inFlight;
-  }
+export type PiscinaWorker = {
+  id: string;
+  currentUsage: number;
+  isRunningAbortableTask: boolean;
+  [kWorkerData]: WorkerInfo;
+  // TODO: maybe add histogram data here?
```
jerome-benoit: Not a maybe, it's a requirement to implement load balancing algorithm.

src/index.ts
```diff
@@ -379,38 +388,111 @@ class ThreadPool {
...
+        if (task.abortSignal) {
+          // console.log('into skipqueue');
+          this.skipQueue.push(task);
+        } else {
+          this.taskQueue.push(task);
+        }
+
+        return false;
+      }
+      case 1: { // add
```
jerome-benoit: It's not the LB code that must take such a decision. It's only responsibility is to pick up a worker given custom criteria among the ones available if any. Worker creation is orthogonal to load balancing (and expected to be done before invoking the LB). 
Furthermore, it will make writing custom LB more complex and error prone than it should. 
Untangling the worker creation code from the LB one a prerequisite.

src/worker_pool/balancer/index.ts
```diff
@@ -0,0 +1,43 @@
...
+      if (
+        !task.isAbortable &&
+        (worker.currentUsage < checkpoint)
+      ) {
+        command.candidate = worker;
+        checkpoint = worker.currentUsage;
+      }
+    }
+
+    return command;
```
jerome-benoit: The array index is usually enough. There's race condition left if the worker crash: the LB API shall include a method to handle worker removal. 
The API needs also to include a method to update the LB internal that will be called at the end of each task execution. LB algos might maintain internal statistics to take decision. The method handling worker removal must then also remove its internal stats relative to it.

src/worker_pool/base.ts
```diff
@@ -0,0 +1,106 @@
...
+      }
+    }
+
+    isReady () : boolean {
+      return this.onreadyListeners === null;
+    }
+
+    onReady (fn : () => void) {
+      if (this.onreadyListeners === null) {
+        fn(); // Zalgo is okay here.
```
jerome-benoit, line 22:
```suggestion
        fn(); // algo is okay here.
```

src/worker_pool/balancer/index.ts
metcoder95: I was assessing wether or not to provide that detailed information; currently my goal was to provide the overall usage (max capacity, tasks running, etc.) to help the LB make decisions.

IMO it provides enough for simple algorithms like RB, but for more complex ones based on usage over time, it will require further data.

src/index.ts
metcoder95: Hmm, this is an interesting point; maybe I saw it in another way.
My initial idea was to help the balancer hint the pool wether to instantiate a new worker or not; but overall, if the balancer decides not to pick one, the pool can make the further decision of spawning and enqueuing.

That can make it cleaner.

src/worker_pool/balancer/index.ts
metcoder95: So reduce the communication semantics to just the _id_ of the worker, so the pool and lb can react to it.

This can be interesting, an `onClose` or `onDestroy` can be called from the pool to hear whenever a worker is down so it can update its internal statistics.

Do you have something specific in mind?

src/worker_pool/base.ts
metcoder95: Its different, the term [Zalgo](https://progressivecoder.substack.com/p/dont-unleash-zalgo-in-your-nodejs) applies here.

src/index.ts
jerome-benoit: Since you're targeting a public API, I really think it should be as simple as possible, like the custom queue one. 

Regarding the worker creation usage, you can add optional tunables with sensible defaults to the LB API to allow to give hints to the pool behavior, such as the direct usage of any newly created worker instead of running the LB algo with it added.  

Something like: 
```ts
{ 
  useCreatedWorker: boolean,
  ...
}
```

src/worker_pool/balancer/index.ts
jerome-benoit: An interesting stat is the average tasks runtime  per worker. A custom LB might implement the algo to pick up the worker with the minimum. 
But it will need: 
- per worker histogram for runtime stats
- sensible default stats runtime at worker creation to avoid biased decision in LB algo

The median runtime is also interesting and the LB code will need to compute it since histogram does not compute it. 

Using the `worker_threads` worker id is better than using the array id but at a cost of doing a lookup to find the worker instance. The array index usage has also drawbacks: the code must be very careful at worker removal and  async task handling, the index can stay valid while not pointing to the correct worker. 

Piscina runtime calculation is also lacking some accuracy: it's not exactly the task execution time on the worker, it also includes worker implementation overhead  + message channel latency.

src/worker_pool/balancer/index.ts
jerome-benoit: @ronag is contacting `worker_threads` pool maintainers one after the other to ask for load balancing features :)

Since I've already spent some times on them in poolifier, without per worker usage stats, designing useful LB will be impossible.

src/index.ts
metcoder95: That would be definitely something interesting to explore, but maybe we can start simply and state that if the balancer returns `null`, Piscina will attempt to enqueue the task or start a new worker.

Further we can explore how a balancer and scheduler can collaborate together; but maybe for simplicity just keep it as is

src/worker_pool/balancer/index.ts
metcoder95: The point you made at the latter was one of the reasons I've had some concerns exposing histograms. The time measured is including the message channel overhead + some extra here and there; it will require some tuning to make it fully valuable for something sensitive to precision as an LB; for overall metric gathering can be enough tho.

>Using the worker_threads worker id is better than using the array id but at a cost of doing a lookup to find the worker instance. 

Overall a `Map` might not impose too much overhead, but we can explore; we can start with that and maybe tune for performance afterwards.

jerome-benoit: > The time measured is including the message channel overhead + some extra here and there; it will require some tuning to make it fully valuable for something sensitive to precision as an LB; for overall metric gathering can be enough tho.

Doing accurate runtime measurement will require to change the current worker API: the measurement needs to be done on the worker side. I've not been able to come up with a better solution than extending an existing worker class. 

> 
> > Using the worker_threads worker id is better than using the array id but at a cost of doing a lookup to find the worker instance.
> 
> Overall a `Map` might not impose too much overhead, but we can explore; we can start with that and maybe tune for performance afterwards.

Been there. And finally I've reverted to using an array: 

- `.reduce()` was faster than iteration on `Map` to consolidate any information from the workers such as finding a min. 
- the number of records in the array will stay low, so `Map` will probably remain overkill

src/index.ts
jerome-benoit: I do not think it's a good idea to expect the worker creation decision logic to be part of the LB code. It will end by having the same code in all LBs: 

Is there a worker available? -> Yes -> Run the LB algo | -> No -> (Is pool full -> Yes -> End | -> No -> Ask for worker creation) 
The LB code will need to be called twice.  

That logic can be factored out the LB code and makes LB implementation easier.

jerome-benoit: > Further we can explore how a balancer and scheduler can collaborate together; but maybe for simplicity just keep it as is

Another point: please version the public API since the beginning. The extra work worth it, it will evolve since it's not like a queue API, it's more like a plugin API.

src/worker_pool/balancer/index.ts
metcoder95: > Doing accurate runtime measurement will require to change the current worker API: the measurement needs to be done on the worker side. I've not been able to come up with a better solution than extending an existing worker class.

That should be fine, the result comes packed with some extra data from the Worker, we can extend that schema to add that calculation.

> Been there. And finally I've reverted to using an array:

Good catch.
Overall sticking to an array and leaving the LB to decide should be fine, any optimization can come afterwards

src/index.ts
metcoder95: > The LB code will need to be called twice.

Hmm, this is interesting. It will require to split the handling of a new worker from when a worker finishes a task, which is currently coupled. That will need to be handled differently.

In that way, upon creation we deliver the first task in the queue; but I imagined the LB, although called twice, might have some benefits if the algorithm is based on some affinity. 

>The extra work worth it, it will evolve since it's not like a queue API, it's more like a plugin API.

Yeah, that's the intention. Thanks for the callout

jerome-benoit: > > The LB code will need to be called twice.
> 
> Hmm, this is interesting. It will require to split the handling of a new worker from when a worker finishes a task, which is currently coupled. That will need to be handled differently.

I guess the code entanglement for worker creation logic and load balancing logic can be part of another PR. I would do it prior to LB API integration. If I remember correctly, in piscina, at task end (message channel response):
- lock is taken on the shared tasks queue 
- LB is run to find an available worker with a minimum of executing tasks, if none available, worker is created if pool not full
- lock is released

I'm not sure about the locking, it's probably taken only if an eligible worker is found. Whatever, testing for workers availability and readyness and create one if needed logic can be factored out and run once in the current code as an optimization. And will help for that PR.

src/worker_pool/balancer/index.ts
```diff
@@ -0,0 +1,43 @@
+import type { PiscinaTask } from '../../task_queue';
+import type { PiscinaWorker } from '..';
+
+export type PiscinaLoadBalancerCommand = {
+  candidate?: PiscinaWorker | null; // If candidate is passed, it will be used as the result of the load balancer and ingore the command
+  command?: 0 | 1; // 1 = add, 0 = busy
+};
+export type PisicnaLoadBalancer = (
+  task: PiscinaTask,
```
jerome-benoit, line 9: Non ready workers must be removed from the array supplied to the LB.

src/worker_pool/balancer/index.ts
jerome-benoit: While thinking of it: non ready workers must be removed from the array supplied to the LB to avoid code to test for it on each LB implementation. Is `this.workers` safe regarding worker readyness?

Plenty of works for the months to come :)

src/index.ts
metcoder95: Yeah, it is not really locking (but mostly a side-effect). Piscina infers that if you already have a task queue that hasn't been emptied yet, new tasks are enqueued so this ensures the tasks are solved in the same way they entered (unless a custom task queue has been set).

Nevertheless, the thread is stopped until a further task is picked.

That sounds good, think that gonna make the implementation easier and move forward.

src/worker_pool/balancer/index.ts
metcoder95: >While thinking of it: non ready workers must be removed from the array supplied to the LB to avoid code to test for it on each LB implementation. Is this.workers safe regarding worker readyness?

Partially, the `Pool` categorizes the workers between ready and non-ready, but when iterating it goes through all of them instead of just the ready ones.

Room for improvement for sure 

>Plenty of works for the months to come :)

Definitely, but plenty of good ideas and optimizations are being made here; thanks for the insightful discussion!

metcoder95:
I've already the list of thing to add to this PR.
To avoid make it so big and also do not close it, I'll start creating PRs to this branch with smaller enhancements here and there.

I'll expedite new `major` from Piscina next week to remove support for older Node.js versions and deprecate some APIs already.

Will work on this over the upcoming months

metcoder95:
If someone is willing to open a PR tackle this work, happy to have a look 👍

src/worker_pool/balancer/index.ts
```diff
@@ -0,0 +1,39 @@
+import type { PiscinaTask } from '../../task_queue';
+import type { PiscinaWorker } from '..';
+
+export type PiscinaTaskBalancer = (
+  task: PiscinaTask,
+  workers: PiscinaWorker[]
```
metcoder95, line 6: I'm doubtful about wether or not pass the workers on every call or make the balancer rely directly on the events.

By passing them on every call, balancer will always have the up-to-date list of workers, but might cause them to need to constantly adapt and lack possibly the state.

On the event based approach they get definitely more control; maybe keep both?

metcoder95:
@ronag @jerome-benoit this should be ready for a review for landing; so far the balancer has been decoupled from an scheduler.

I'll update docs and more later this week

jerome-benoit:
> @ronag @jerome-benoit this should be ready for a review for landing; so far the balancer has been decoupled from an scheduler.

Great!

I do not think I will be able to do a review before next week ...

metcoder95:
No rush! I'll proceed with the documentation this week

src/index.ts
```diff
@@ -369,39 +399,81 @@ class ThreadPool {
...
-      this._maybeDrain();
-      return;
+
+      const distributed = this._distributeTask(taskInfo, workers);
+
+      // If task was distributed, we should continue to distribute more tasks
+      if (distributed) {
+        continue;
+        // If balancer states that pool is busy, we should stop trying to distribute tasks
+      } else { break; }
```
```suggestion
      if (distributed) {
        // If task was distributed, we should continue to distribute more tasks
        continue;
      } else { 
        // If balancer states that pool is busy, we should stop trying to distribute tasks
        break; 
      }
```

src/index.ts
```diff
@@ -369,39 +399,81 @@ class ThreadPool {
...
+      return true;
+    }
+
+    // We spawn if possible
+    // TODO: scheduler will intercept this.
+    if (this.workers.size < this.options.maxThreads) {
+      this._addNewWorker();
+    }
+
+    if (task.abortSignal) {
```
ronag, line 469:
```suggestion
    if (task.abortSignal?.aborted) {
```
?

ronag (commented):
RSLGTM. A bit too much for me to get a good overview. Would be great with some docs and samples. I think sorting out the public API is higher prio than the implementation details.

src/index.ts
metcoder95: We do that check already in upper layers, if it reaches here, it means the task is not aborted yet.

metcoder95:
SGTM, I'll start preparing either Tomorrow or later in the weekend

src/task_queue/index.ts
```diff
@@ -87,12 +89,25 @@ export class TaskInfo extends AsyncResource implements Task {
 
       this.filename = filename;
       this.name = name;
+      // TODO: This should not be global
```
jerome-benoit, line 92: And it should a UUID to avoid overflow.

src/worker_pool/balancer/index.ts
jerome-benoit: You can't guarantee a consistent list of workers because of the async nature of the event loop, even as an class attribute passed as an arguments: worker removal is depending on an event ...

One way that is still racy in some corner cases is the use of a flag on the worker information that is set as soon as the worker is planned for removal.

src/worker_pool/base.ts
```diff
@@ -0,0 +1,140 @@
...
+      this.onTaskDoneListeners[i](item);
+    }
+  }
+
+  onTaskDone (fn : (item : T) => void) {
+    this.onTaskDoneListeners.push(fn);
+  }
+
+  getCurrentUsage (): number {
+    let inFlight = 0;
```
jerome-benoit, line 117: tasks currently executed?

src/index.ts
```diff
@@ -369,39 +399,83 @@ class ThreadPool {
...
+      candidate[kWorkerData].postTask(task);
+      this._maybeDrain();
+      // If candidate, let's try to distribute more tasks
+      return true;
+    }
+
+    // We spawn if possible
+    // TODO: scheduler will intercept this.
+    if (this.workers.size < this.options.maxThreads) {
+      this._addNewWorker();
```
jerome-benoit: How is the worker histogram initialized in that case? 
If a balancer query the runtime and that is initialized to zero, the balancer will wrongly pickup only new workers. 
The same goes for worker recreation in the error case.

src/index.ts
```diff
@@ -424,9 +498,9 @@ class ThreadPool {
     filename = maybeFileURLToPath(filename);
 
     let signal: AbortSignalAny | null;
-    if (this.closingUp) {
+    if (this.closingUp || this.destroying) {
```
jerome-benoit, line 500: What is the difference between closingUp and destroying flag? I think the two can be merged.

src/index.ts
```diff
@@ -369,39 +399,83 @@ class ThreadPool {
...
+      this.waitTime?.record(toHistogramIntegerNano(now - task.created));
+      task.started = now;
+      candidate[kWorkerData].postTask(task);
+      this._maybeDrain();
+      // If candidate, let's try to distribute more tasks
+      return true;
+    }
+
+    // We spawn if possible
+    // TODO: scheduler will intercept this.
```
jerome-benoit: What do you mean here? I think worker creation should be done before calling the balancer: the newly created worker is flagged ready and usable.

src/index.ts
```diff
@@ -369,39 +399,83 @@ class ThreadPool {
...
+    if (this.workers.size < this.options.maxThreads) {
```
jerome-benoit: That condition looks incorrect to me: the pool start with the minimum size will trigger a worker creation a each task distribution. 

The semantic of a pool busy needing a worker creation is: all workers are running concurrently their task quota and pool max size is not reached.  And it seems that check must be done in the load balancer code. The two (worker choice and worker creation) are orthogonal and implementing it that way will make the LB API less complicated.

src/worker.ts
```diff
@@ -156,7 +158,8 @@ function onMessage (
       response = {
         taskId,
         result: result,
-        error: null
+        error: null,
+        time: start == null ? null : Number(process.hrtime.bigint() - start)
```
jerome-benoit:
```suggestion
        time: start == null ? null : performance.now() - start)
```

src/worker.ts
```diff
@@ -143,6 +143,8 @@ function onMessage (
   (async function () {
     let response : ResponseMessage;
     let transferList : any[] = [];
+    const start = message.histogramEnabled === 1 ? process.hrtime.bigint() : null;
```
jerome-benoit:
```suggestion
    const start = message.histogramEnabled === 1 ? performance.now() : null;
```

Using `process.hrtime.bigint()` has a cost. And I doubt that cost worth it for LB decision.

src/worker.ts
```diff
@@ -175,7 +178,8 @@ function onMessage (
         result: null,
         // It may be worth taking a look at the error cloning algorithm we
         // use in Node.js core here, it's quite a bit more flexible
-        error: <Error>error
+        error: <Error>error,
+        time: start == null ? null : Number(process.hrtime.bigint() - start)
```
jerome-benoit:
```suggestion
        time: start == null ? null : performance.now() - start)
```

src/worker_pool/base.ts
```diff
@@ -0,0 +1,140 @@
...
+      const usage = item.currentUsage();
+      if (usage === 0) return item;
+      if (usage < minUsage) {
+        candidate = item;
+        minUsage = usage;
+      }
+    }
+
+    return candidate;
+  }
```
jerome-benoit: That code should be moved to the default load balancer function and use the workers args: worker load balancing is meant to be totally untangled from the rest.

test/workers.ts
```diff
@@ -0,0 +1,56 @@
...
+        workersSecondRound = workers;
+      } else if (
+        workersFirstRound[0].id !== workers[0].id
+      ) {
+        workersSecondRound = workers;
+      }
+      // Verify distribution to properly test this feature
+      const candidate = workers[index++ % workers.length];
+
+      if (candidate.currentUsage !== 0 && !candidate.isRunningAbortableTask) {
```
jerome-benoit, line 31: It's not the purpose of the load balancer to test if the worker chosen is usable or not: it should be factored out. Otherwise LBs will duplicate that code again and again: the main code ask for a worker to be chosen, that's all. If the worker is not eligible (busy, not available, aborted), the main code see if a new worker can be created and ask again. 
That does not hinder the LB to also decide to integrate availability checks, but minimum LB requirement code is to chose a worker. Lean requirement.

src/worker_pool/index.ts
```diff
@@ -167,6 +95,10 @@ export class WorkerInfo extends AsynchronouslyCreatedResource {
     }
 
     _handleResponse (message : ResponseMessage) : void {
+      if (message.time != null) {
+        this.histogram?.record(Math.max(1, message.time));
```
jerome-benoit: I do not understand the Math.max: if the time-stamping can give a zero execution time, there's an issue with it.

src/task_queue/index.ts
metcoder95: Interesting, do you see a need for universal uniqueness?

My first idea was to provide a simple counter that limits to the SMI optimization and re-starts; maybe in combination with the Worker ID; and if need for enabling more complex use-cases ask the caller to provide an ID generator

src/worker_pool/base.ts
metcoder95: Yes 👍

src/index.ts
metcoder95: Good call

Overall, the current state is to just spawn new workers based on balancer feedback.

How it right now works is:
- Based on `minThreads` and `maxThreads` Piscina will spawn the necessary workers
- On every task distributed or enqueued, we will check `maxThreads` to verify if we can create more workers

In the future, how I envision this is:
- Based on `minThreads` and `maxThreads` Piscina will spawn the necessary workers
- When a task is about to be distributed
  - Ask the scheduler to verify if we should spawn more workers (and spawn if so)
  - Call the balancer to distribute the workload
  - If not possible to distribute, we enqueue

src/index.ts
metcoder95: Hmm, good catch; I see what you mean.

This should be decoupled to not complicate the implementation. These should be handled directly before or after the distribution, but should have nothing to do with the balancing.

Let me fix this

src/index.ts
metcoder95: For each new worker, the histogram is instantiated to zero, but the whole WorkerPool has a global histogram that can be queried.

Nevertheless, to query it, you need to have a reference to the Worker Pool (Piscina instance). Maybe worth it to pass a reference to the global Piscina instance so it can query this data

src/worker.ts
metcoder95: Agree, let me also see if we can take advantage and abstain of using `hrtime` in favour of `performance.now` on the places where we query this.

jerome-benoit:
Le dimanche 13 octobre 2024 à 04:11 -0700, Carlos Fuentes a écrit :
> Interesting, do you see a need for universal uniqueness?
I think piscina already account for executed tasks via histogram.
Basically you can use a simple counter but you have to make sure
overflowing is properly handled. A web spider running for several
months might overflow it. 

-- 
Jérôme Benoit aka fraggle
Piment Noir - https://piment-noir.org
OpenPGP Key ID : 27B535D3
Key fingerprint : B799 BBF6 8EC8 911B B8D7 CDBC C3B1 92C6 27B5 35D3

src/worker_pool/balancer/index.ts
metcoder95: Do you mean something like `toDestroy: true | false` (or similar)?

src/worker_pool/base.ts
metcoder95: Good catch, this is a leftover mostly; I'll clean this up

src/worker_pool/index.ts
metcoder95: I noticed this from a previous issue, where the problem was non-deterministic, couldn't reproduce it easily.

I added it as a workaround but need to dig deeper into it

test/workers.ts
metcoder95: Get your point, but what are you referring to exactly on this?

To bring context, this was boilerplate to test the workers were hinted correctly that the workers were marked as destroyed correctly; true I copy-pasted from other code (that's why agree with your statement).

metcoder95:
Thanks for the feedback @jerome-benoit, I'll kick-off the second round along the week

src/task_queue/index.ts
jerome-benoit: High rate tasks submission workload could overflow it if no restart is done since months. Isn't the task execution counting already done via histogram?
Can it also overflow?

src/index.ts
jerome-benoit: Maybe. Or you can initialize to the pool one if it's not a cumulative histogram.
On poolifier, each entries init to the min of worker matching entries.

src/worker_pool/base.ts
jerome-benoit: As learning examples for devs: 
- the reimplementation of the current LB using the new API.
- And also a round robbin one: very low overhead but still efficient under load.

test/workers.ts
jerome-benoit: Isn't the conditions testing for the eligibility of the chosen worker? If not, that review comment is wrong. 

The intent is the same as a previous comment: once LB has chosen a worker, the worker eligibility check and/or worker creation conditions check and/or LB retry should be done outside the LB code.

src/worker_pool/index.ts
jerome-benoit: With the high resolution timestamping at ns?

src/index.ts
jerome-benoit: The previous code assumed if no available worker were chosen, that was meaning all workers were executing their task quota -> pool not full -> worker creation.
Decoupling the two in the LB will end up to a simplier to explain, understand and implement LBs.

src/task_queue/index.ts
metcoder95: True, but do we would need to track down every task individually for more than 30 days? I imagine that maybe you'll like it for debugging purposes, but not sure how much after that.

For that, caller should implement a custom solution that fits its use case; I'd like to provide a good defaults with less overhead and allow extensibility.

>Isn't the task execution counting already done via histogram?
Can it also overflow?

Noup. I imagine the task id just for debugging purposes, beyond that I don't see much of a use for it.

src/index.ts
metcoder95: That would be useful, I'll follow that path; thanks!

test/workers.ts
metcoder95: >Isn't the conditions testing for the eligibility of the chosen worker?

Not really, it just aims to test the workers are marked as `destroyed` whenever they get destroyed; used that _distribution implementation_ for simplicity and copy paste from others

src/worker_pool/index.ts
metcoder95: Yeah, got issue reports on that and I also was able to reproduce it from time to time, but was not deterministic. Every N times could reproduce it.

In theory, when having `0` out of the high ms resolution with `performance.now`, means start of the process, so I imagine goes in that sense.

jerome-benoit: Sounds like a bug to me. But dunno where it came from.

src/worker_pool/balancer/index.ts
jerome-benoit: The existing `destroying` flag should be enough. But the code need be audited to ensure: 
- set as soon as a worker is planned for removal
- checked in the right code paths

leesoryqq (approved):

metcoder95:
@ronag I've updated the docs, feel free to take a look now 👍

jerome-benoit:
> @ronag I've updated the docs, feel free to take a look now 👍

If you are sure that no regressions are introduced by that PR, you should merge it and flag load balancer feature as experimental. Then do an alpha or beta or ... release and communicate about it to allow people to play with it and do feedbacks.

metcoder95:
Agree, that's the plan. After this is merged, I'll start issuing `alpha` of the new version of the library to start gathering feedback and allow people use it.

I'll reflect this in the documentation of this PR

metcoder95:
I'm merging now, if there's more feedback please write it here, I'll be iterating over it in upcoming PRs

Let's land and release some alpha release of the new version to get some feedback; once the Atomics implementation is also merged, I'll release the version.

PR #620, opened 2024-07-26
feat: add state properties to worker

metcoder95:


src/worker_pool/index.ts
```diff
@@ -124,6 +136,7 @@ export class WorkerInfo extends AsynchronouslyCreatedResource {
     }
 
     processPendingMessages () {
+      if (this.terminating || this.destroyed) return;
```
metcoder95:
```suggestion
      if (this.destroyed) return;
```

test/workers.ts
```diff
@@ -0,0 +1,61 @@
+import { resolve } from 'node:path';
+
+import { test } from 'tap';
+
+import Piscina from '../dist';
+
+test('workers are marked as detroyed if destroyed', async t => {
```
ronag:
```suggestion
test('workers are marked as destroyed if destroyed', async t => {
```

src/worker_pool/index.ts
```diff
@@ -156,13 +169,25 @@ export class WorkerInfo extends AsynchronouslyCreatedResource {
     get interface (): PiscinaWorker {
       const parent = this;
       return {
-        id: this.worker.threadId.toString(),
-        currentUsage: this.currentUsage(),
-        isRunningAbortableTask: this.isRunningAbortableTask(),
+        get id () {
+          return parent.worker.threadId.toString();
```
ronag: Why to string?

ronag (approved):

src/worker_pool/index.ts
metcoder95: Its a number by default, was wondering if shall we keep it like this or better parse it as string

ronag: I would suggest to keep it as a number. More efficient to use than a string.

PR #625, opened 2024-08-04
feat: add worker events to pool

metcoder95:


metcoder95:
if there's extra comments, feel free to leave them and I'll tackle it in the feature branch

Issue #615, opened 2024-07-18
Can you add an example to README as to how to broadcast a message to all worker threads using `BroadcastChannel` (Node v18+)?


titanism:
Per discussions in [these closed issues](https://github.com/search?q=repo%3Apiscinajs%2Fpiscina+send+message&type=issues), it's clear `piscina` users want to be able to broadcast messages to worker threads.

Could you please add an example to the README as to how to broadcast a message to all threads (and listen in the thread for the message)?

metcoder95:
Closed by #656

PR #656, opened 2024-09-15
docs: An example that implements broadcast communication

testwhygh:
It's a PR to solve the problem "[add an example to README as to how to broadcast a message to all worker threads using BroadcastChannel](https://github.com/piscinajs/piscina/issues/615)". This code has been tested,

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,45 +268,46 @@ In `main.js`
 ```js
 'use strict';
 
-const { BroadcastChannel } = require("worker_threads");
+const { BroadcastChannel } = require('worker_threads');
 const { resolve } = require('path');
 
 const Piscina = require('piscina');
 const piscina = new Piscina({
-    filename: resolve(__dirname, 'worker.js'),
-    useAtomics: false
+  filename: resolve(__dirname, 'worker.js'),
+  useAtomics: false
 });
 
-async function main() {
-    const bc = new BroadcastChannel('my_channel');
-    // start worker
-    Promise.all([
-        piscina.run("thread 1"),
-        piscina.run("thread 2")
-    ]);
-    // post message in one second
-    setTimeout(() => {
-        bc.postMessage('Main thread message');
-    }, 1000);
+async function main () {
+  const bc = new BroadcastChannel('my_channel');
+  // start worker
+  Promise.all([
+    piscina.run('thread 1'),
+    piscina.run('thread 2')
+  ]);
+  // post message in one second
+  setTimeout(() => {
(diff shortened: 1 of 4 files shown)


metcoder95 (commented):
This is amazing, thanks!
Could you rather add it to the `docs/examples` section?

ronag (commented):
You should also make sure to set the flag that makes sure piscina doesn't use blocking Atomics.wait

testwhygh:
> This is amazing, thanks! Could you rather add it to the `docs/examples` section?

Of course, I will also add it to `docs/docs/examples`

testwhygh:
> You should also make sure to set the flag that makes sure piscina doesn't use blocking Atomics.wait

Yes, setting the flag `useAtomics` to false avoids using Atomics.wait, but why should I do this? Is it for better concurrency performance?

metcoder95:
> > You should also make sure to set the flag that makes sure piscina doesn't use blocking Atomics.wait
> 
> Yes, setting the flag `useAtomics` to false avoids using Atomics.wait, but why should I do this? Is it for better concurrency performance?

Is to avoid the thread to be blocked when idle (we use `Atomic.wait` to put the thread on hold until further tasks are enqueued).

testwhygh:
> > > You should also make sure to set the flag that makes sure piscina doesn't use blocking Atomics.wait
> > 
> > 
> > Yes, setting the flag `useAtomics` to false avoids using Atomics.wait, but why should I do this? Is it for better concurrency performance?
> 
> Is to avoid the thread to be blocked when idle (we use `Atomic.wait` to put the thread on hold until further tasks are enqueued).

Thanks, I have modified it.

metcoder95 (dismissed):
lgtm

testwhygh:
> lgtm
Hello, should I do anything else to get this PR completed?

README.md
```diff
@@ -264,6 +264,56 @@ module.exports = ({ a, b }) => {
...
+```js
+'use strict';
+
+const { BroadcastChannel } = require('worker_threads');
+const { resolve } = require('path');
+
+const Piscina = require('piscina');
+const piscina = new Piscina({
+  filename: resolve(__dirname, 'worker.js'),
+  useAtomics: false
```
metcoder95, line 281: Let's just add a comment of why `useAtomics` is disabled, and it should be ready to go

testwhygh: Of course, it is necessary

metcoder95 (approved):

PR #687, opened 2024-10-30
feat: support Atomics.waitAsync

### LINKED ISSUE #346: use `Atomic.waitAsync` instead of `Atomics.wait`
In order to not fully block the thread and allow background stuff to continue running.

metcoder95:
- **feat: Support Atomics.waitAsync**


metcoder95:
Adding test-case is pending

src/worker.ts
```diff
@@ -122,7 +124,17 @@ function atomicsWaitLoop (port : MessagePort, sharedBuffer : Int32Array) {
     // Check whether there are new messages by testing whether the current
     // number of requests posted by the parent thread matches the number of
     // requests received.
-    Atomics.wait(sharedBuffer, kRequestCountField, lastSeenRequestCount);
+    
+    if (useAsyncAtomics === true) {
+      // @ts-expect-error
+      const { async, value } =  Atomics.waitAsync(sharedBuffer, kRequestCountField, lastSeenRequestCount);
+
+      return async === true && value;
```
ronag: Where is this awaited?

metcoder95: [Here](https://github.com/piscinajs/piscina/pull/687/files#diff-1245a5bbc4ca0684dff978efcae118b08f62d2afc05812df22d0627985b4273aR88) and [here](https://github.com/piscinajs/piscina/pull/687/files#diff-1245a5bbc4ca0684dff978efcae118b08f62d2afc05812df22d0627985b4273aR155).

They're a IIFE that has a `catch` to handle the errors; it is not awaited explicitly to avoid wrapping a non-promise into a Promise, nevertheless we might be loosing some stack out of the stacktrace. although as it is executed in a single place it might not be that bad but happy to hear your thoughts

ronag: I still don't see how it's awaited... Maybe to complicated for me to understand

metcoder95: Maybe I didn't get what you meant at the first time, but think I might get it now. At a top-level is not really awaited, but more on a fire-and-forget approach, so the Promise is not awaited for resolution but triggered and any error is handled.

This happens on every new message or at a startup

ronag: That doesn't sound good? Though I don't know enough of the current design to truly know.

metcoder95: Yeah, while writing this, I noticed the same; let me see if there's any side-effect on awaiting on it.
I don't see a potential problem as anyways we already await for the task handler to resolve the task, so it should come for free; but let me check and benchmark

metcoder95: Refactored into awaiting for the wake up; although the benchmarks requires some re-work, the refactor reflected a nice performance boost

src/worker.ts
```diff
@@ -122,78 +130,102 @@ function atomicsWaitLoop (port : MessagePort, sharedBuffer : Int32Array) {
...
+    // before potentially entering the `Atomics.wait()` loop, and before
+    // returning the result so that messages will always be printed even
+    // if the process would otherwise be ready to exit.
+    if (process.stdout.writableLength > 0) {
+      await new Promise((resolve) => process.stdout.write('', resolve));
     }
-    currentTasks--;
+    if (process.stderr.writableLength > 0) {
+      await new Promise((resolve) => process.stderr.write('', resolve));
+    }
```
ronag: In async mode you don't need to worry about this. Skip it.

ronag (approved):

src/worker.ts
ronag: Would be nice to resolve this but not strictly necessary.

README.md
```diff
@@ -362,12 +362,16 @@ This class extends [`EventEmitter`][] from Node.js.
...
-    If `useAtomics` is `true`, it will cause to pause threads (stoping all execution)
-    between tasks. Ideally, threads should wait for all operations to finish before 
-    returning control to the main thread (avoid having open handles within a thread).
+    setting the environment variable `PISCINA_DISABLE_ATOMICS` to `1` .
+    If `atomics` is `sync`, it will cause to pause threads (stoping all execution)
+    between tasks. Ideally, threads should wait for all operations to finish before
+    returning control to the main thread (avoid having open handles within a thread). If still want to have the possibility
+    of having open handles or handle asynchrnous tasks, you can set the environment variable `PISCINA_ENABLE_ASYNC_ATOMICS` to `1` or setting `options.atomics` to `async`.
+
+   > **Note**: The `async` mode comes with performance penalties and can lead to undesired behaviour if open handles are not tracked correctly.
```
ronag, line 373: Can you explain the open handle thing? Why is that an issue with async?

ronag (dismissed):

ronag:
I would recommend making async the default for a future semver major.  IMHO it's the best balance between  perf and expected behavior.

README.md
metcoder95: Mostly as we are not doing event loop ticks, these can add some overhead if the workers start doing more than it should.

Even waiting for just the worker to wake up added some ticks that showed a difference of ±5% which is really negligible.

I've not strong opinion in this one; but if we want to make this the default behaviour we should document that it is important to not keep the workers doing too much background tasks as they can face some penalty (as well if using `async`, implementer needs to account for the teardown of the worker when idle, i.e. no more tasks scheduled)

metcoder95: If that sounds good, I can adjust the PR for that 👍

metcoder95:
Merging now, let me know if there's further feedback so I can address it in further PRs

jdmarshall:
So with this change will setInterval work properly?  When this PR was first brought to my attention I was concerned in part about whether something like OpenTelemetry.js would function properly in a piscina worker thread. 

A feature I'm thinking about creating a PR for would make piscina more attractive for beefier use cases, but the absence of ticks when no unanswered messages are in flight is problematic.

metcoder95:
Yeah; in a brief, with this change the event loop of the thread will continue ticking instead of being hold as it was before.
This can be adjusted through the `atomics` flag when instantiating Piscina or `PISCINA_ENABLE_ASYNC_ATOMICS`.

jdmarshall:
Excellent.  

As an aside, I'm wondering... A couple years ago I had a process that wouldn't exit. There's a debugging library that can tell you why the event loop isn't exiting in NodeJS. I think it hooks setInterval and setTimeout so it can dump the metadata about everything that is keeping things alive.

I'm wondering if it would be saner for piscina to do something similar, and keep the event loop in the workers running as long as there are outstanding intervals or timeouts. The bigger problem I'm trying to solve:

What if I try to push a task that's big enough to have OpenTelemetry or Prometheus running in it into a worker thread? Its Interval will never fire and the last block of telemetry after each message will either disappear or show up potentially minutes later. 

Because without that, this other line of questioning (making 2 pools of workers talk to each other) is dead in the water. Because that sort of long-lived worker definitely needs to have monitoring. The flag is workable but it becomes an asterisk for things like this, and one or two are okay but beyond that I find you start to lose your audience.

metcoder95:
I believe that it should be totally doable with this new change; you can use the current alpha version for experimental testing and play with it to see how it goes.

PR #723, opened 2025-01-05
feat(#305)!: Expose new `PiscinaHistogram` abstraction

### LINKED ISSUE #305: feat: allow to reset histograms
I'm collecting `runTime` and `waitTime` metrics for my pool with otel:

```ts
const metrics = {
  workersQueueWaitTime: meter.createObservableGauge('content_workers_queue_wait_time'),
  workersQueueLatency: meter.createObservableGauge('content_workers_queue_latency')
} as const

// later on
const percentiles = ['p50', 'p99', 'p99_9'] as const
metrics.workersQueueWaitTime.addCallback((result) => {
  for (const percentile of percentiles) {
    result.observe(this.workers.waitTime[percentile], { percentile })
  }
})
metrics.workersQueueLatency.addCallback((result) => {
  for (const percentile of percentiles) {
    result.observe(this.workers.runTime[percentile], { percentile })
  }
})
```

However over the lifetime of the pool, those histogram are never reset so the p99 and p99_9 take a lot of time to get back to a current value since it stores old one. I would like to be able to call `this.workers.runTime.reset()` to clear the buckets after each value collection.

If that's something you are okay with, i can make a PR for it

metcoder95:
Closes #305


ronag (dismissed):

ronag:
Should we automatically reset the histogram at some interval?

docs/docs/api-reference/class.md
```diff
@@ -110,6 +110,53 @@ Use caution when setting resource limits. Setting limits that are too low may
 result in the `Piscina` worker threads being unusable.
 :::
 
+## `PiscinaHistogram`
+
+The `PiscinaHistogram` allows you to access the histogram data for the pool of worker threads.
```
RafaelGSS, line 115: Can we have a code sample on how users would do it in userland?

RafaelGSS:
> Should we automatically reset the histogram at some interval?

I don't think so, since we are storing distribution data (mean, avg, stddev)

docs/docs/api-reference/class.md
metcoder95: Done, let me know if that's enough or if there's a something else to cover

RafaelGSS (approved):
LGTM.

I think the PR title doesn't express entirely what this PR does. It also exposes the PiscinaHistogram

PR #757, opened 2025-03-13
feat: Allow long-running threads

### LINKED ISSUE #678: Is there a way in which we can create long living threads?
My request may sound strange, but bare with me, currently piscina is designed for worker threads is more functional stateless approach and this may work for 90% use cases, but my use case requires me to hold a instance of class on a thread and keep it long running. The reason for this is cause i hold iterator on the worker thread side and exchange bulk messages for iteration between it and main thread.. can this be done via piscina?

What i am doing is opening sqlite DBs queries on worker threads in iterator mode and doing a k-way merge on main thread.. and i dont want to keep reopening these queries.. any suggestions?

lucianolix:
This PR aims to allow long running threads in Piscina, by leveraging the `Infinity `value on the `idleTimeout` option (See issue #678)

Using this value avoids the `setTimeout` which removes the worker when the thread is idle from being created.


metcoder95 (approved):

metcoder95 (commented):
Can you just add to the docs a warning stating that, if `Infinity` is used, it can lead to over use and waste of resources?

PR #782, opened 2025-05-02
chore!: drop v18

metcoder95:


dasa:
Should this be updated too? 

https://github.com/piscinajs/piscina/blob/c8421adb47ba5c93f7387db688639c2006715c6a/package.json#L12-L14

metcoder95:
Good call, sure! Do you want to open a PR for that?

dasa:
> Do you want to open a PR for that?

Created PR #786

Issue #612, opened 2024-07-14
chore: Migrate tests from Tap to `node:test`


metcoder95:
Given the current state of `node:test` as testing library and runner, we can start using it in combination with `borp` to run our tests.

pmarchini:
Hey @metcoder95, I've opened a PR to address this 😁

metcoder95:
Closed by #792

PR #792, opened 2025-05-21
chore: migrate tests from tap to node:test

pmarchini:
This should address #612!

I added a small utility script to support different Node versions and operating systems.
I didn’t use `borp` because I encountered errors similar to those reported in this issue: [https://github.com/mcollina/borp/issues/43](https://github.com/mcollina/borp/issues/43).

Alternative solutions are definitely welcome 😁


scripts/run-tests.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+// As node 20 test runner does not support glob patterns in input
+// and considering that we could have multiple OS we manually
+// resolve the test files and pass them to the test runner
+
+const { spawnSync } = require('child_process');
+const { globSync } = require('glob');
+
+const testFiles = globSync('test/**/*test.ts', { absolute: true });
```
metcoder95: This lgtm, the only proposal I'd like to add is to receive the pattern from the argv instead, so we have the chance to reuse this implementation across other use cases.

Maybe even the flags, but we can start with the glob pattern first

metcoder95:
thanks a lot for the contribution!

scripts/run-tests.js
pmarchini: sure 😁 I'll change it asap 🚀

pmarchini: Pushed!

On another matter: I was checking the output of the coverage and it seems different.
I've been checking the results and, as far as I can tell, they seem correct even though they are different.

Could you please take a look too?

pmarchini: This is an example of the new output of `coverage/lcov-report/src/histogram.ts.html`
image

metcoder95: Interesting, the actual coverage shown by your PR is more accurate than the one we currently have.

The uncovered lines are correct, can we address that before landing this?

pmarchini: sure! I'll work on this as soon as I can!

pmarchini: Hey @metcoder95, I tried taking a closer look at the uncovered lines, but I'm not confident about adding tests as I can't ensure their quality due to lacking deep knowledge of the codebase and its implementation.

I might end up adding superficial coverage that doesn't provide proper protection against regressions!

I think it would be better if someone with more experience took care of it!!!

metcoder95: Sure thing, please open an issue and we can take it from there 👍

metcoder95 (approved):

PR #801, opened 2025-06-03
feat: make piscina entrypoint easier for typescript

danielpza:
This is technically a breaking change in the typescript api. 

Changed the generics from

```ts
const pool = new Piscina<Parameters<typeof import("./worker.js").default>[0], ReturnType<typeof import("./worker.js").default>>({ filename: "./worker.js" })

// or specified manually
const pool = new Piscina<{a: number, b: number}, number>({ filename: "./worker.js" })
```

to

```ts
const pool = new Piscina<typeof import("./worker.js")>({ filename: "./worker.js" })

// or specified manually
const pool = new Piscina<{default: (payload: { a: number, b: number }) => number}>({ filename: "./worker.js" })
```

Also works with multiple exports:

```ts
// worker.ts
export function greet({ name }: { name: string }): string {
  return `Hello ${name}!`;
}

export default function add({ a, b }: { a: number; b: number }): number {
  return a + b;
}
```

```ts
// main.ts
const pool = new Piscina<typeof import("./worker.js")>({ filename: "./worker.js" })

pool.run({
  // types are checked for default export
}) // returns Promise<number>

pool.run({
  // types are checked for greet export
}, { name: 'greet' }) // returns Promise<string>
```

This (imo) improves the typescript usage by reducing the amount of code needed to properly type the worker function.


metcoder95 (commented):
I really like this, but sadly it imposes a breaking change in the current version; can we aim to add a fallback to the current behavior if the generic passed is not an object?

danielpza:
> I really like this, but sadly it imposes a breaking change in the current version; can we aim to add a fallback to the current behavior if the generic passed is not an object?

I played around with the code a bit to see if it was possible, unfortunately I think it's not possible (or I don't know enough TS) to make it backwards compatible.

metcoder95:
I'm thinking on [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads), but not sure if they can be applied to classes 🤔 

Otherwise, we can point this feature directly to v6, but will take some time to land as it will be aimed to next year.

danielpza:
> I'm thinking on [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads), but not sure if they can be applied to classes 🤔

I thought the same but couldn't get it to work, seems it's not possible.

> Otherwise, we can point this feature directly to v6, but will take some time to land as it will be aimed to next year.

I'm fine with this going to v6. 

If someone later finds a way to backport it without breaking compatibility that would be great.

metcoder95:
Perfect, then let's do this. I'll sync `next` branch to contain latest changes and we point this PR to that branch instead.

If willing, we can give it a try to backport it aftewards, otherwise we can open an Issue for someone that might be interested in contribute

danielpza:
@metcoder95 rebased to the current branch. Can you approve the workflows run again?

metcoder95:
CI failures on Windows are expected; not related to you PR

test/types-entrypoint.ts
```diff
@@ -0,0 +1,31 @@
+import { resolve } from "node:path";
+import { expectTypeOf } from "expect-type";
+
+import { test } from "tap";
```
metcoder95: Let's remove `tap` we are not using tap anymore

test/types-entrypoint.ts
```diff
@@ -0,0 +1,31 @@
...
+
+test("types can be inferred from function", async (t) => {
+  const pool = new Piscina<typeof import("./fixtures/worker-type.js")>({
+    filename: resolve(__dirname, "fixtures/worker-type.js"),
+    concurrentTasksPerWorker: 1,
+  });
+
+  expectTypeOf(pool.run).toExtend<(payload: { a: number; b: number }) => Promise<number>>();
+  expectTypeOf(pool.run).toExtend<(payload: { a: number; b: number }, opts: { name: "default" }) => Promise<number>>();
+  expectTypeOf(pool.run).toExtend<(payload: { name: string }, opts: { name: "greet" }) => Promise<string>>();
```
metcoder95: I'd like to split these kind of testing as, although useful, it sometimes can be messy; thinking on keeping the `.ts` files for testing functionality and adding `test.d.ts` for actually testing types.

danielpza: > adding test.d.ts for actually testing types.

I think we cannot use expectTypeOf in declaration only files. However I renamed it to `types.test-d.ts` modified the runner to exclude `*.test-d.ts` files similar to https://vitest.dev/guide/testing-types.html. See https://github.com/piscinajs/piscina/pull/801/commits/d197f88f676b5e66c5a6ab53067e7b0c6f33d1c8

test/types-entrypoint.ts
danielpza: Fixed in https://github.com/piscinajs/piscina/pull/801/commits/801f82dc40c385fa31978321216a69ccaeb3e036

test/types-entrypoint.ts
metcoder95: can we add a npm script that tests it?

danielpza: done in https://github.com/piscinajs/piscina/pull/801/commits/6cf8f019fc53107a040e040d9ec646ae619de5ce, covered in the lint script

metcoder95 (approved):

metcoder95:
Test seems to fail :(

danielpza:
> Test seems to fail :(

Pushed another commit, this should fix it now.

metcoder95:
Seems like it still doesn't like it

danielpza:
> Seems like it still doesn't like it

~I messed up the steps order~, I fixed it, it should be good now.

metcoder95 (approved):

metcoder95:
Thanks for the contribution!

PR #807, opened 2025-06-12
fix(#805): Fix handling of aborted tasks

### LINKED ISSUE #805: Later tasks never start if next task is aborted
We are using piscina to run one task at a time (`minThreads = maxThreads = 1`). We noticed an unexpected behavior when tasks are aborted in a certain order (details below). The tasks themselves contain a long sleep, so you can assume that task 1 is still running when it is aborted, whereas task 2 and task 3 have not started yet.

1. submit task 1
2. submit task 2
3. submit task 3
4. abort task 2
5. abort task 1

We would expect the following trace, which is in fact what we get in piscina 4.9.2:
```
$ node main.js
Started 1
task2 rejected with "AbortError: The task has been aborted"
task1 rejected with "AbortError: The task has been aborted"
Started 3
Finished 3
$
```

However, in piscina 5.0.0, task 3 never starts and the program exits:
```
$ node main.js
Started 1
task2 rejected with "AbortError: The task has been aborted"
task1 rejected with "AbortError: The task has been aborted"
$
```

Here is the code that produced the traces above:

worker.js:
```js
const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));

module.exports = async ({ x }) => {
    console.log("Started " + x);
    await sleep(10000);
    console.log("Finished " + x);
};
```

main.js:
```js
"use strict";

const Piscina = require("piscina");
const { resolve } = require("path");

const piscina = new Piscina({
    filename: resolve(__dirname, "worker.js"),
    maxThreads: 1,
    minThreads: 1
});

const abortController1 = new AbortController();
const abortController2 = new AbortController();
const abortController3 = new AbortController();

// Use timeouts to make sure that tasks end up in the correct order in the queue
setTimeout(() => {
    piscina.run(
        { x: 1 },
        { signal: abortController1.signal }
    ).catch(reason => console.log("task1 rejected with \"" + reason + "\""));
}, 1000);

setTimeout(() => {
    piscina.run(
        { x: 2 },
        { signal: abortController2.signal }
    ).catch(reason => console.log("task2 rejected with \"" + reason + "\""));
}, 2000);

setTimeout(() => {
    piscina.run(
        { x: 3 },
        { signal: abortController3.signal }
    ).catch(reason => console.log("task3 rejected with \"" + reason + "\""));
}, 3000);

setTimeout(() => {
    abortController2.abort()
}, 4000);

setTimeout(() => {
    abortController1.abort()
}, 5000);
```

We suspect this has something to do with the interaction between `_onWorkerAvailable` and `_distributeTask`. `_distributeTask` returns `false` if a task is aborted, and `_onWorkerAvailable`, upon failing to distribute a task, breaks out of its loop and does not consider the rest of the tasks in the queue:
https://github.com/piscinajs/piscina/blob/8b285fa50dbe35969b181755df70d689bf3c9a83/src/index.ts#L422-L435

metcoder95:
Fixes #805

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -38,7 +38,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node-version: [20.x, 22.x, 23.x]
+        node-version: [20.x, 22.x, 24.x]
     runs-on: ${{matrix.os}}
     steps:
       - uses: actions/checkout@v4
--- a/docs/docs/api-reference/class.md
+++ b/docs/docs/api-reference/class.md
@@ -114,6 +114,14 @@ Use caution when setting resource limits. Setting limits that are too low may
 result in the `Piscina` worker threads being unusable.
 :::
 
+**  :::info
+    **Note on Explicit Resource Management**: Piscina does has support for `Symbol.dispose` and `Symbol.asyncDispose` for explicit resource management for its usage with the `using` keyword.
+    This is only avaiable on Node.js 24 and higher.
+
+    For more information, see the [Explicit Resource Management](https://github.com/tc39/proposal-explicit-resource-management).
+    :::
+**
+
 ## `PiscinaHistogram`
 
 The `PiscinaHistogram` allows you to access the histogram data for the pool of worker threads.
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -4503,9 +4503,10 @@
       }
     },
     "node_modules/brace-expansion": {
-      "version": "1.1.11",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
-      "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
+      "version": "1.1.12",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.12.tgz",
+      "integrity": "sha512-9T9UjW3r0UW5c1Q7GTwllptXwhvYmEzFhzMfZ9H7FQWt+uZePjZPjBP/W1ZEyZ1twGWom5/56TF4lPcqjnDHcg==",
+      "license": "MIT",
       "dependencies": {
         "balanced-match": "^1.0.0",
         "concat-map": "0.0.1"
--- a/src/histogram.ts
+++ b/src/histogram.ts
@@ -51,10 +51,6 @@ export class PiscinaHistogramHandler {
     return this.#runTime.count;
   }
 
-  get waitTimeCount(): number {
-    return this.#waitTime.count;
-  }
-
   recordRunTime(value: number) {
     this.#runTime.record(PiscinaHistogramHandler.toHistogramIntegerNano(value));
   }
--- a/src/index.ts
+++ b/src/index.ts
@@ -845,6 +845,14 @@ export default class Piscina<T = any, R = any> extends EventEmitterAsyncResource
     return this.#pool.destroy();
   }
 
+  [Symbol.dispose]() {
+    this.close();
+  }
+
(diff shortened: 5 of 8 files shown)


mattdailis:
I tested this change in my project, and it does appear to fix the issue :+1: I wasn't sure what is the most effective way to get your new code into my project, so I manually edited my `node_modules/piscina/dist/index.js` file with your change.

I do have a couple more questions:
- When we paused the debugger in `_onWorkerAvailable`, we found that all of the tasks were in `this.skipQueue`, and that `this.taskQueue` was always empty. Is that expected in this case? We thought that was surprising given the comment "The skipQueue will have tasks that we previously shifted off the task queue but had to skip over", which seems to imply we would have seen them in the taskQueue at some point. Perhaps we missed that point when debugging
- I cloned the piscina repo and copied your new test (but not the fix) onto the `current` branch, and ran it, expecting it to fail. It did not fail for me - were you able to get it to fail?
  - I wasn't sure how to run the test properly, so I added a new script to `package.json` with the following: `"test:mattdailis": "node scripts/run-tests.js --coverage --pattern='test/pool.test.ts'",`

metcoder95:
>When we paused the debugger in _onWorkerAvailable, we found that all of the tasks were in this.skipQueue, and that this.taskQueue was always empty. Is that expected in this case? 

Yes, it is expected. The `skipQueue` holds tasks that wants to be prioritised as they are abortable. So in order to isolate them, Piscina will attempt to flush them to the first available worker asap so the time to abort is close the actual caller expectation.

Otherwise, it pushed into the queue, is highly possible that during high concurrent tasks, the tasks gets aborted before even being attempted to be executed.

>I cloned the piscina repo and copied your new test (but not the fix) onto the current branch, and ran it, expecting it to fail. It did not fail for me - were you able to get it to fail?

Good callout, just figured out the same now; let me adjust it.
But you ran the tests correctly 👍

metcoder95:
I've been spending more time understand why the tests also returns green on `current`, and this roots to mostly a timing situation.

Piscina, once a task has been aborted, tears down the worker that was executing the task and spins up a new one; on a single time script, the timing is short that this does not allow Piscina to properly allow to wait for a further worker to be available, leading to the task not running at all.

Nevertheless, with the test runner this timing exists.
Unless we find another way to properly shorten this and test it, this is the best effort to attempt cover the case at hand

mattdailis:
> I've been spending more time understand why the tests also returns green on `current`, and this roots to mostly a timing situation.
> 
> Piscina, once a task has been aborted, tears down the worker that was executing the task and spins up a new one; on a single time script, the timing is short that this does not allow Piscina to properly allow to wait for a further worker to be available, leading to the task not running at all.
> 
> Nevertheless, with the test runner this timing exists. Unless we find another way to properly shorten this and test it, this is the best effort to attempt cover the case at hand

Understood. Thanks so much for the quick turnaround!

vilacha:
> > When we paused the debugger in _onWorkerAvailable, we found that all of the tasks were in this.skipQueue, and that this.taskQueue was always empty. Is that expected in this case?
> 
> Yes, it is expected. The `skipQueue` holds tasks that wants to be prioritised as they are abortable. So in order to isolate them, Piscina will attempt to flush them to the first available worker asap so the time to abort is close the actual caller expectation.
> 
> Otherwise, it pushed into the queue, is highly possible that during high concurrent tasks, the tasks gets aborted before even being attempted to be executed.
> 
> > I cloned the piscina repo and copied your new test (but not the fix) onto the current branch, and ran it, expecting it to fail. It did not fail for me - were you able to get it to fail?
> 
> Good callout, just figured out the same now; let me adjust it. But you ran the tests correctly 👍

I don't understand the "skipQueue" logic. Let's say that I want to have all my tasks to be "abortable". If I add a signal to all instantiations of  a task, none of them goes into TaskQueue, they all are inside "skipQueue". Looking into the source code, one can see that aborting a task kills the thread when needed or removes the given task from TaskQueue, which is what the user would expect. To me skipQueue seems to be unnecessary. Abortable tasks should not have a greater priority. I think this behaviour should be reconsidered. As it is right now it is imposible to build a queue with cancellable tasks, only possible when returning at least 1 in the size method of a custom queue.

metcoder95:
Can you open a new issue to redirect the conversation there?

PR #817, opened 2025-06-17
Change assert.strictEqual to a safe early exit

bryanrideshark:
There is a `assert.strictEqual(workerInfo.currentUsage(), 0);` that multiple organizations report, breaks the Angular 20 CLI.

Exiting early by throwing an error is unnecessary and only causes programs relying on piscina to have non deterministic failure states.

This pull request removes the assert, instead of throwing an error, an early return is the appropriate response.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -443,7 +443,7 @@ class ThreadPool {
     // If more workers than minThreads, we can remove idle workers
     if (workerInfo.currentUsage() === 0 &&
         this.workers.size > this.options.minThreads) {
-        workerInfo.idleTimeout = setTimeout(() => {        
+      workerInfo.idleTimeout = setTimeout(() => {        
         if (workerInfo.currentUsage() !== 0) {
           // Exit early - we can't safely remove the worker.
           return;
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,7 +4,6 @@ import { resolve } from 'node:path';
 import { inspect, types } from 'node:util';
 import { performance } from 'node:perf_hooks';
 import { setTimeout as sleep } from 'node:timers/promises';
-import assert from 'node:assert';
 
 import { version } from '../package.json';
 import type {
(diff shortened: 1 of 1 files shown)


metcoder95 (commented):
The issue seems to be more of a symptom of another bug; I’m under the impression that `idleTimeout` is not being cleared when a new task is send to a worker.

I’ll check out this in deep.

bryanrideshark:
Yes, it does seem like this error only surfaced because of the clearIdleTimeout not being reliably called in another method, there seems to be another pull request that rectifies that problem.

I just think it's not a good idea to have a footgun like this hanging around, especially since the intent of this method seems to be a nice, automatic clean up. Other developers in the future could easily re-introduce the problem by failing to clear the idle timeout for some reason, and this change would make it safe forever, without changing the important logic.

bryanrideshark:
Seems like the tests failed because I didn't remove the import - That change has been made, you can run them again and I think they'd pass.

metcoder95:
~~Yeah, let's flush this and release it asap; I'll work on the proper fix after that and release a second patch~~

Let's narrow our focus to #818

metcoder95 (approved):

src/index.ts
```diff
@@ -444,7 +443,10 @@ class ThreadPool {
     if (workerInfo.currentUsage() === 0 &&
         this.workers.size > this.options.minThreads) {
       workerInfo.idleTimeout = setTimeout(() => {
-        assert.strictEqual(workerInfo.currentUsage(), 0);
+        if (workerInfo.currentUsage() !== 0) {
+          // Exit early - we can't safely remove the worker.
+          return;
+        }
```
metcoder95, line 449: Can you expand the documentation here a bit?
Meaning that, due to race conditions, we cannot exit as worker stills in usage?

metcoder95 (commented):
After evaluating deeply and seeking other options, it seems this is the best one.

So far, it seems with the addition of the LoadBalancer abstraction, race conditions are a bit prone if the `idleTimeout` are close to the actual wall-time it takes for a task to resolve and distribute new ones.

PR #818, opened 2025-06-17
fix: prevent race condition in idle worker cleanup

### LINKED ISSUE #816: `workerInfo.idleTimeout` not cleared in time, leading to `AssertionError: 1 !== 0 ` since version 4.9.0
When a worker thread is used after `workerInfo.idleTimeout` has been set, the `idleTimeout` is not cleared before `currentUsage` is increased. This results in an [AssertionError](https://github.com/piscinajs/piscina/blob/fe781d66040eb87446692a3f7056c9c2c7b84a4a/src/index.ts#L447), as the system expects the worker's `currentUsage()` to be 0 when the idleTimeout fires, but it is actually 1 (or more) because the worker is active.

**Stack Trace:**
```
/home/runner/_work/clickup_frontend/clickup_frontend/node_modules/.pnpm/piscina@5.0.0/node_modules/piscina/src/index.ts:447
        assert.strictEqual(workerInfo.currentUsage(), 0);
               ^
AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
1 !== 0
    at Timeout._onTimeout (/home/runner/_work/clickup_frontend/clickup_frontend/node_modules/.pnpm/piscina@5.0.0/node_modules/piscina/src/index.ts:447:16)
    at listOnTimeout (node:internal/timers:588:17)
    at processTimers (node:internal/timers:523:7) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: 1,
  expected: 0,
  operator: 'strictEqual'
}
```
See: https://github.com/angular/angular-cli/issues/30549

alan-agius4:
Resolves an issue where workers could be terminated while still in use due to an ordering problem between increasing worker usage and clearing the idle timeout. This caused `AssertionError: 1 !== 0` during cleanup.

Closes #816

### DIFF
--- a/src/worker_pool/index.ts
+++ b/src/worker_pool/index.ts
@@ -121,6 +121,7 @@ export class WorkerInfo extends AsynchronouslyCreatedResource {
       };
 
       try {
+        this.clearIdleTimeout();
         this.port.postMessage(message, taskInfo.transferList);
       } catch (err) {
         // This would mostly happen if e.g. message contains unserializable data
@@ -132,7 +133,6 @@ export class WorkerInfo extends AsynchronouslyCreatedResource {
       taskInfo.workerInfo = this;
       this.taskInfos.set(taskInfo.taskId, taskInfo);
       this.ref();
-      this.clearIdleTimeout();
 
       // Inform the worker that there are new messages posted, and wake it up
       // if it is waiting for one.


metcoder95 (commented):
can you add a test for it?
Just to facilitate detect regressions

alan-agius4:
@metcoder95 I considered adding a test, but I'm not sure how to detect this specific case to ensure the timeout gets cleared earlier. Happy to include it if you have any suggestions!

src/worker_pool/index.ts
```diff
@@ -111,6 +111,7 @@ export class WorkerInfo extends AsynchronouslyCreatedResource {
     postTask (taskInfo : TaskInfo) {
       assert(!this.taskInfos.has(taskInfo.taskId));
       assert(!this.terminating && !this.destroyed);
+      this.clearIdleTimeout();
```
metcoder95: Let's move this within the `try/catch` before posting the message to the port.

metcoder95 (commented):
Is a bit harder to get a good reproduction; here i'm assuming the problem roots to a high event loop lag that delay the execution of the timer callbacks leading to a race condition between the execution and the new task being posted.

src/worker_pool/index.ts
alan-agius4: Moved.

metcoder95 (approved):

Tallyb:
still getting it (building storybook with angular): 
```
AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
1 !== 0
    at Timeout._onTimeout (.../node_modules/piscina/dist/index.js:317:39)
    at listOnTimeout (node:internal/timers:588:17)
    at process.processTimers (node:internal/timers:523:7) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: 1,
  expected: 0,
  operator: 'strictEqual'
}
Node.js v22.16.0
```

PR #819, opened 2025-06-19
test: simplify

AugustinMauroy:
## Description

- Reduce sleeping time for some test
- Remove such as possible type casting
- Use assert module instead of test connect assertion wich is uncommon way to make test

## Result

- **Before** 23s
- **After with test coverage** 12s
- **After without test coverage** 9s

> That on my MacBook M1 on node v20

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,6 @@
     "lint": "eslint",
     "test": "node scripts/run-tests.js --pattern='test/**/*test.ts'",
     "test:ci": "npm run lint && npm run build && npm run test:coverage",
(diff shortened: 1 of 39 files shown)


metcoder95 (commented):
lgtm. Just tests seems red a bit

AugustinMauroy:
The problem was easier to fix than expected. It was just that the test timeout was too close to the one in the worker, so the worker hadn't necessarily finished its work when the test wait finished.

metcoder95 (approved):

PR #823, opened 2025-06-24
test: remove `c8`

AugustinMauroy:


AugustinMauroy:
how i had tested localy for 22 but not for 20. Why node is doing that ...

metcoder95 (approved):

metcoder95:
It is strange, the issue seems to be only GH Actions, I cannot reproduce it locally

metcoder95:
Seems the issue is related to source maps, haven't isolated it to try to reproduce it standalone; feel free to give it a shot if you have time: https://github.com/nodejs/node/pull/53315

AugustinMauroy:
@metcoder95 using `--enable-source-maps` didn't work so I just don't collect coverage for v20 if we have one coverage it's should be same for the rest of other version

metcoder95 (approved):

metcoder95 (approved):

PR #828, opened 2025-06-25
refactor: enqueue clear idle timeout to micro task queue

metcoder95:


metcoder95:
@alan-agius4 can you verify if this reliefs the issue a bit

src/worker_pool/index.ts
```diff
@@ -132,6 +131,7 @@ export class WorkerInfo extends AsynchronouslyCreatedResource {
 
       taskInfo.workerInfo = this;
       this.taskInfos.set(taskInfo.taskId, taskInfo);
+      queueMicrotask(() => this.clearIdleTimeout())
```
alan-agius4, line 134: Would this increase the occurrences, the clear timeout happens after the task is added, meaning that if there is an existing timeout running, the assertion could fail.

alan-agius4:
> @alan-agius4 can you verify if this reliefs the issue a bit

Sure, let me check later during the day.

src/worker_pool/index.ts
metcoder95: It can be, but I'm not 100% sure that will be the case.
By adding this to the `queueMicrotask`, the enqueued callback should be executed between every event loop tick, meaning that if the event-loop is about to enter the timers phase, `queueMicrotask` gets flushed beforehand and the callback scheduler at the timers should be cleared up.

Tho, I cannot fully tested without a proper reproduction

alan-agius4:
I tested a bit and I was no able to reproduce with this change, although It was also hard to reproduce with the current release.

metcoder95:
Let me release the change and will be pending on further regressions

mattlewis92:
Hi! Just a heads up, this has not fixed the assertion issue, we are still seeing it throw with the fix in 5.1.2:
```
~/repo/node_modules/.pnpm/piscina@5.1.2/node_modules/piscina/src/index.ts:447
        assert.strictEqual(workerInfo.currentUsage(), 0);
               ^
AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
1 !== 0
    at Timeout._onTimeout (~/repo/node_modules/.pnpm/piscina@5.1.2/node_modules/piscina/src/index.ts:447:16)
    at listOnTimeout (node:internal/timers:588:17)
    at processTimers (node:internal/timers:523:7) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: 1,
  expected: 0,
  operator: 'strictEqual'
}
```

metcoder95:
This is by using the Angular CLI right?

mattlewis92:
> This is by using the Angular CLI right?

Yep! Sorry I'm not able to provide a repro, as the issue seems to be correlated to have a large project running inside of github actions, so attempting to slim things down will fix the race condition. Happy to add debug logs though and post the results here if you can show me where the best places to add them would be?

metcoder95:
If you can bring some information like the `idleTimeout`,  `max/minThreads` and `maxQueue` could be a good starting point.

On top, an idea of what's the task overall latency (you can use [`Piscina#runTime`](https://piscinajs.dev/api-reference/Properties#property-runtime-readonly) for that) will give me a better idea of what could be happening.

mattlewis92:
> If you can bring some information like the `idleTimeout`, `max/minThreads` and `maxQueue` could be a good starting point.
> 
> On top, an idea of what's the task overall latency (you can use [`Piscina#runTime`](https://piscinajs.dev/api-reference/Properties#property-runtime-readonly) for that) will give me a better idea of what could be happening.

Sure thing. I added `console.log('options', JSON.stringify(this.options));` [here](https://github.com/piscinajs/piscina/blob/18e86dfebae214a3fd99fd7f1e7d37a29dc2f36c/src/index.ts#L447) and this was the output:

```
{"filename":"/node_modules/@angular/build/src/tools/sass/worker.js","name":"default","minThreads":1,"maxThreads":8,"idleTimeout":4000,"maxQueue":null,"concurrentTasksPerWorker":1,"atomics":"sync","taskQueue":{"tasks":[]},"niceIncrement":0,"trackUnmanagedFds":true,"closeTimeout":30000,"recordTiming":true,"workerHistogram":false}
```

The `runTime` property seems to be undefined though, I tried all of these:

```
console.log('runTime', this.histogram, this.runTime, this.histogram?.runTime);
```

and it just logged `PiscinaHistogramHandler {} undefined undefined`

metcoder95:
>The runTime property seems to be undefined though, I tried all of these:

Did you run tasks beforehand?

metcoder95:
Can I know as well what is run on these workers? The `idleTimeout` seems healthy, 4s is not a big problem; but i imagine time to submit tasks are taking longer, causing the overlap (not excluding this is a bug)

metcoder95:
oh, btw, `workerHistogram` seems disabled

mattlewis92:
> > The runTime property seems to be undefined though, I tried all of these:
> 
> Did you run tasks beforehand?

> Can I know as well what is run on these workers? The idleTimeout seems healthy, 4s is not a big problem; but i imagine time to submit tasks are taking longer, causing the overlap (not excluding this is a bug)

This is the angular CLI code that runs the worker, it will only create the pool once it tries to create tasks:
https://github.com/angular/angular-cli/blob/6b4512b6b968ac5dfc0745e3b30aee2f7098970f/packages/angular/build/src/tools/sass/sass-service.ts#L139-L154

> oh, btw, workerHistogram seems disabled

After setting `workerHistogram: true`, logging `console.log('runTime', this.histogram, this.runTime, this.histogram?.runTime);` still outputs `runTime PiscinaHistogramHandler {} undefined undefined`

mattlewis92:
After adding some logging around timing i discovered something interesting. [The assertion](https://github.com/piscinajs/piscina/blob/current/src/index.ts#L447) is ran and passes 8 times (maxThreads is set to 8) when the pool is first created. It's only ~27 seconds later that the assertion runs again and fails.

metcoder95:
Hmm, there's a mismatch there; I was also able to reproduce it with our benchmarks, so I've now a better understanding of what might be happening. What i'll have to check is wether we are either removing late the task that has been finished or not clearing the timeout on the right time.

Thanks for all the context!

mattlewis92:
> Hmm, there's a mismatch there; I was also able to reproduce it with our benchmarks, so I've now a better understanding of what might be happening. What i'll have to check is wether we are either removing late the task that has been finished or not clearing the timeout on the right time.
> 
> Thanks for all the context!

I think the timeout not being cleared is the issue here. Adding this patch [here](https://github.com/piscinajs/piscina/blob/70e09395c8ec0129271464b0ff158ab88bf7870f/src/index.ts#L442) fixes the issue:

```
if (workerInfo.idleTimeout) {
	clearTimeout(workerInfo.idleTimeout);
	workerInfo.idleTimeout = undefined;
}
```

PR #837, opened 2025-07-04
refactor: tweaks

metcoder95:
- **refactor: task info done**
- **refactor: tweaks**

### DIFF
--- a/src/worker_pool/base.ts
+++ b/src/worker_pool/base.ts
@@ -43,7 +43,7 @@ export abstract class AsynchronouslyCreatedResource {
     }
 
     isDestroyed () {
-      return this.ondestroyListeners != null;
+      return this.ondestroyListeners == null;
     }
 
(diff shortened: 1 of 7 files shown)


src/worker_pool/base.ts
```diff
@@ -35,15 +35,15 @@ export abstract class AsynchronouslyCreatedResource {
...
+      assert(listeners != null);
       this.ondestroyListeners = null;
       for (const listener of listeners) {
         listener();
       }
     }
 
     isDestroyed () {
-      return this.ondestroyListeners === null;
+      return this.ondestroyListeners != null;
```
metcoder95:
```suggestion
      return this.ondestroyListeners == null;
```

metcoder95:
```suggestion
      return this.ondestroyListeners == null;
```

PR #866, opened 2025-07-29
feat: stream iterators

metcoder95:


metcoder95:
will refactor some implementation around and take it from scratch again

PR #872, opened 2025-08-05
feat!: drain/needsDrain accounts for task queue

### LINKED ISSUE #868: needsDrain is emitted too late
Hi! im currently trying to use piscina to run a large amount of tasks. i was hoping to be able to use the `needsDrain` event to manage and split my tasks into chunks. but the problem i have is that this event is emitted too late for me to catch and wait on. for example, i run 10 tasks, with my pool having `maxQueue: 2` and `maxThreads: 1`. after the third task is passed and queue is full, even at the start of task 4, the `pool.needsDrain` is false. and only after spawning all of my tasks and 7 of them failing, the `needsDrain`  event is emitted. also the `pool.needsDrain` is only set to true after failing one task, but no event is emitted at this point.
Am i doing something wrong? or i should watch for `pool.queueSize` at my code manually?

metcoder95:
Closes #868

### DIFF
--- a/test/abort-task.test.ts
+++ b/test/abort-task.test.ts
@@ -50,7 +50,7 @@ test('tasks can be aborted through EventEmitter before running', async () => {
   const ee = new EventEmitter();
   const task1 = pool.run(bufs[0]);
   const abortable = pool.run(bufs[1], { signal: ee });
-  assert.strictEqual(pool.queueSize, 0); // Means it's running
+  assert.strictEqual(pool.queueSize, 1); // Means it's running and abortable enqueued
   assert.rejects(abortable, /The task has been aborted/);
 
   ee.emit('abort');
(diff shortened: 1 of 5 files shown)


metcoder95:
@ehsanghorbani190 @jdmarshall

test/abort-task.test.ts
```diff
@@ -50,7 +50,7 @@ test('tasks can be aborted through EventEmitter before running', async () => {
   const ee = new EventEmitter();
   const task1 = pool.run(bufs[0]);
   const abortable = pool.run(bufs[1], { signal: ee });
-  assert.strictEqual(pool.queueSize, 0); // Means it's running
```
jdmarshall, line 53: I have a PR against p-limit to reduce the nextTick depth by one and the confirmation that something was off was found in the tests, which were asserting similarly weird numbers for queue length. I just inverted a counter and queue manipulation to eliminate the nextTick call. Ans then had to fix the tests similarly.

ehsanghorbani190:
@metcoder95 Hi again! glad to see this PR is merged! i checked the npm and yet the last publish was two months, is this PR published or just merged?

metcoder95:
Haven't prepared yet the cut-off for the new v6; sadly this is a breaking change so cannot land on current stable :(

Unless done through an flag, which can be an option.

PR #903, opened 2025-10-01
docs: Fix link formatting for Custom Task Queues

missinglink:
docs: Fix link formatting for Custom Task Queues (requires minus delim rather than underscore)


missinglink:
As an aside, the [docs](https://github.com/piscinajs/piscina?tab=readme-ov-file#built-in-queues) mention the `FixedQueue` as having improved performance but doesn't mention any downsides.

Is there any reason I shouldn't use it, is there some reason why it wasn't promoted to the default?

metcoder95 (approved):

metcoder95:
`FixedQueue` is the default task queue from `v5` onwards.

There's no potential downside of using it besides performance reasons. It behaves pretty well under big size workloads (many concurrent tasks), but not as good as `ArrayQueue` in smaller ones.

PR #963, opened 2026-01-28
fix: onWorkerMessage gets skipped in Jest environment 

linnerissa:
Repro: Set up a jest test utilizing a piscina workerpool with min:0 max: N. When the first worker gets set up, it fails to execute the task because the onWorkerMessage check for `message instanceOf Object` fails.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -314,7 +314,14 @@ class ThreadPool {
     }
 
     function onWorkerMessage (this: ThreadPool, message: any) {
-      message instanceof Object && READY in message ? onWorkerReady() : onEventMessage.call(this, message);
+      const isReadyMessage =
+        (message instanceof Object && READY in message) ||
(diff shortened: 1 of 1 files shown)


metcoder95 (commented):
Can we add a regression test for it?

linnerissa:
@metcoder95 https://github.com/piscinajs/piscina/pull/966 check this out for regression test (it adds jest dependencies)

linnerissa:
branch that shows regression tests passing with fix c: 
https://github.com/linnerissa/piscina/pull/1

metcoder95 (commented):
Seems is causing a bit of friction, can you check it out?

linnerissa:
mismanaged my branches resulting in a bad merge conflict. that should be resolved now. ready for workflow rerun

metcoder95 (approved):

PR #966, opened 2026-02-04
test: showing piscina failures in Jest

linnerissa:
Jest doesnt pass the onWorkerReady message check to execute work after new threads are spun up. Added Jest to show that. 

package.json 
- added 3 dev dependencies to support jest tests
- added commands for jest tests 
- altered CI to include the new jest test

native tests need to skip test/jest subdirectory
added a native version of the new jest test to explicitly force reliance on adding a new thread to pass

### DIFF
--- a/test-jest/ready-message.test.ts
+++ b/test-jest/ready-message.test.ts
@@ -1,0 +1,30 @@
+import { resolve } from 'node:path';
+import Piscina from '../dist';
+
+jest.setTimeout(5000);
+
+describe('Piscina ready message detection', () => {
+  let pool: Piscina | null = null;
+
+  afterEach(async () => {
+    if (pool) {
+      await pool.destroy();
+      pool = null;
+    }
+  });
+
+  it('should detect worker ready message with minThreads=0', async () => {
+    pool = new Piscina({
+      filename: resolve(__dirname, '../test/fixtures/eval.js'),
+      minThreads: 0,
+      maxThreads: 1
+    });
+
+    expect(pool.threads.length).toBe(0);
+
+    const result = await pool.run('42');
+
+    expect(result).toBe(42);
+    expect(pool.threads.length).toBe(1);
+  });
+});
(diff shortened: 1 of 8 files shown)


metcoder95 (commented):
Thanks for the contribution! This is super helpful

Can you just add it to the `.github/workflows/node.yml` as part of the CI?

linnerissa:
@metcoder95 i added the test to `test:ci`. the workflow automatically picks this up in its `run tests` step. 

this approach should be better for local - anyone running `test:ci` will also run the jest tests. less likely to be surprised by the new jest suite @ pr stage.

metcoder95:
Tests seems failing

linnerissa:
scripts/run-tests.js ignore parameter actually resolves diff in macos vs windows. changed approach 🙏

test-jest/ready-message.test.ts
```diff
@@ -0,0 +1,30 @@
+import { resolve } from 'node:path';
```
metcoder95, line 1: I believe something is going wrong with the test on windows, but the output on console doesn't seem that helpful.

linnerissa: can we try rerunning the test? i think it was due to the jest tests being inside the test directory, and being triggered by the node native test runs (which fails due to jest) 

macos and windows resolve file paths differently making it hard to filter out the jest directory without specifying both /jest and \jest

metcoder95:
Hmm, I'm wondering if the file resolution is being the culprit here

linnerissa:
OK i've actually solved it this time - verified on my forked repo's workflow runs

previously the `"test:coverage": "node scripts/run-tests.js --coverage --pattern='test/**/*test.ts'",` quoting on the pattern was resulting in 0 paths found in windows (unlike ubuntu and macos)  which defaults to autodiscovery. this would always capture the jest tests. 

by renaming the jest tests (.spec.ts) and removing the pattern filtering all together, we cleanly get two camps of tests. the existing suite and the new jest suite.

metcoder95 (commented):
Amazing work! Can you resolve conflicts?
After that, it should be ready to go

metcoder95 (approved):

PR #977, opened 2026-02-09
Exclude test files from npm package

veksen:
## Summary

Adds `test` to the existing `.npmignore` to exclude the `test/` directory from the published npm package.

The test files (~208KB) are shipped to npm but serve no purpose for consumers. With ~7M weekly downloads, excluding them reduces aggregate bandwidth.

## Changes

- Added `test` to `.npmignore`

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -11,4 +11,4 @@ output.txt
 tsconfig.json
 eslint.config.js
 .editorconfig
-.worktrees
+.worktreestest


.npmignore
```diff
@@ -11,4 +11,4 @@ output.txt
 tsconfig.json
 eslint.config.js
 .editorconfig
-.worktrees
\ No newline at end of file
+.worktreestest
```
metcoder95, line 14: new line missed?

PR #1032, opened 2026-04-27
feat: ensure task scheduling fairness and FIFO order

### LINKED ISSUE #1031: Queue fairness problem
## Description

When the queue is constantly busy, the tasks are shifted from the `taskQueue` or `skipQueue` and then pushed into the end of one of these queues, breaking the FIFO logic.

In my app, I have a single-threaded worker pool as a server, and 20 clients post their tasks in a loop. In real life, it's a Node.js app and K6 test runner, but I have made a synthetic reproduction example and it shows the same behavior - see the code below.

 - 20 clients keep sending their requests for 1 minute
 - some of the early requests keep being postponed
 - after 1 minute, the incoming task flow ends
 - the postponed tasks are finally executed, showing incredibly high waiting time (dozens of seconds, up to 60 seconds)

The `waitTime` histogram shows the maximum of 60s

## Reproduction
```js
import Piscina from "piscina";
import pLimit from "p-limit";

const pool = new Piscina({
    filename: new URL("./worker.js", import.meta.url).href,
    minThreads: 1,
    maxThreads: 1
});

const limiter = pLimit(20);
const end = Date.now() + 60000;
for (let i = 0; Date.now() < end; i++) {
    if (limiter.activeCount >= limiter.concurrency) {
        i--;
        await new Promise(setImmediate);
        continue;
    }
    const start = Date.now();
    limiter(() => pool.run({i, start})).then(
        () => {
            const end = Date.now();
            console.log(`[${i}] ${start} - ${end} (${end - start}ms)`);
        }
    );
}

await new Promise(r => setTimeout(r, 3000));

console.log(pool.histogram.runTime);
console.log(pool.histogram.waitTime);
```

`worker.js`:
```js
export default function() {
    const d = Date.now();
    while (Date.now() - d < 10) { /* busy wait */ }
}
```

## Environment
- MacOS 26.4.1
- Node 20.19.6
- Piscina 5.1.4

## Expected Behavior

Each task should wait for 19 other tasks to complete (19 * 10ms = 190ms), then spend 10ms running. So each log line should show something around ~200ms

## Actual Behavior

Newer tasks take priority over older ones. As a result, each task takes ~110ms to complete. On my machine, one of tasks `2, 4, 6, 8` is completed at the end of the test and reports the execution time of 60s

kokovtsev:
## Overview

Fixes #1031 

**Note:** I was unsure about the target branch for this change (v5 vs current), please let me know if the PR should be recreated to target a different branch.

In some code paths, a task is shifted from the tail of the queue, but then if we failed to distribute it to a worker it is pushed to the _head_ of the queue.

This breaks the FIFO ordering of the tasks, which is not a big problem by itself, because the real order of execution is non-deterministic in a multi-threaded environment. However, this leads to the fairness issues, where some tasks are waiting in queue more that others, causing latency jitter. In some edge cases (I have consistently observed this when using 1 worker), the same task may lose its place in the queue numerous times, getting delayed by dozens of seconds.

Implementation details:

1. `_distributeTask` is not responsible for returning a task back to the queue anymore. Now the code calling `_distributeTask` should return the task into a correct place (head or tail, skipQueue or taskQueue)
2. A new operation `unshift` has been added to the `TaskQueue` interface and the 2 builtin implementations. This operation is made optional, so that clients who provide custom implementations of `TaskQueue` can continue using them without needing to write new code.

## Category

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation
- [ ] Maintenance (refactoring, dependency updates, etc.)

## Contribution Checklist

- [x] I have reviewed the [guidelines](../CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](../CODE_OF_CONDUCT.md).
- [x] My code adheres to the project's style and conventions.
- [x] I have added tests to cover my changes.
- [x] All tests passed locally.
- [ ] I have updated relevant documentation.

## DCO Sign-off

- [x] I certify that I have the right to submit this contribution under the open source license ([DCO 1.1](../CONTRIBUTING.md#developers-certificate-of-origin-11)).


metcoder95 (commented):
Thanks for the contribution! LGTM

The only concern I've is the change on the current behavior. Although we provide backwards compatibility for custom `task_queues` that does not provides the `unshift` method, the PR per-se changes the current behaviour of the queues that implementations might already have some expectations on.

Not sure if we should guard this through a flag or similar, just to preserve the current behavior and allow opt in for the new one.

kokovtsev:
Good point, thanks!

I can add a new `stricterFIFO` flag for enabling this change, does that sound right for you?

metcoder95:
Sure, let's do that so it can also land on `v5` 👍

metcoder95:
I'm gonna go ahead and send it to v6, feel free to open a pr againg `v5` branch to back port it

metcoder95:
lint seems failing

PR #1042, opened 2026-05-21
fix: include skipQueue tasks in queueSize getter (backport to v5)

### LINKED ISSUE #1029: queueSize is incorrect when using AbortSignal
## Description
pool.queueSize is incorrect when tasks are submitted with a signal. In that case, tasks that cannot be scheduled immediately may end up in skipQueue, but the queueSize getter only accounts for taskQueue. As a result, queued tasks in skipQueue are not reflected in pool.queueSize, which can remain 0 even though tasks are clearly waiting.

## Reproduction

```js
const piscina = require('piscina');

const pool = new Piscina({
  filename: './worker.js',
  minThreads: 1,
  maxThreads: 1,
});

const controller = new AbortController();

const promises = []
for (let i = 0; i < 10; i++) {
  const promise = pool.run({
    reqId,
    timestamp: Date.now()
  }, { signal: controller.signal });
  
  console.log(`[REQ ${i.toString().padStart(3)}] queueSize: ${pool.queueSize}, threads: ${pool.threads.length}, completed: ${pool.completed}`);
  promises.push(promise);
}

await Promise.all(promises);


// worker.js
module.exports = async ({ reqId }) => {
  await new Promise((resolve) => setTimeout(resolve, 2000));
  return reqId;
};
```

With signal the queueSize is always 0. Without signal the queueSize is correct.

## Environment
- **OS**: Windows 11
- **Node.js version**: 22.14.0
- **Piscina version**: 5.1.4

## Expected Behavior

queueSize should reflect the number of tasks currently waiting to be executed, regardless of whether those tasks were submitted with an AbortSignal or not.

algojogacor:
This is an independent contribution and is not associated with any hackathon or competition.

## Summary

Backport the queueSize fix from the `current` branch to `v5`. The queueSize getter now accounts for tasks in skipQueue that were submitted with an AbortSignal.

## Root Cause

When tasks are submitted with an AbortSignal and cannot be scheduled immediately, they are placed in the `skipQueue` (line 468 of src/index.ts) instead of the `taskQueue`. However, the `queueSize` getter only counted `taskQueue.size`, ignoring tasks waiting in `skipQueue`. This caused `pool.queueSize` to report 0 even when tasks were queued.

## Fix

Added `+ pool.skipQueue.length` to the queueSize getter calculation:

```
- return Math.max(pool.taskQueue.size - pool.pendingCapacity(), 0);
+ return Math.max((pool.taskQueue.size + pool.skipQueue.length) - pool.pendingCapacity(), 0);
```

This is a backport of the fix already present on the `current` branch.

## Testing

- The existing `skipQueue` integration already drains during `_maybeDrain()` — the getter now reflects its contents.
- No behavioral regression: the returned value only increases (or stays the same) since we added, never subtracted.

Closes #1029


metcoder95 (commented):
can you add a test for it?

PR #1100, opened 2026-07-10
feat: add stricterFIFO option for queue fairness

### LINKED ISSUE #1031: Queue fairness problem

huntiezz:
Closes #1031

## Problem

When the pool couldn't dispatch a task, `_distributeTask()` pushed it back onto the **tail** of the queue after it had already been shifted off the **head**. Under sustained load (especially with one worker), the same task could keep getting bumped back and wait much longer than tasks submitted later.

## Fix

Add an opt-in `stricterFIFO` pool option (default `false`, so existing behavior is unchanged).

When enabled:

- Undispatched tasks go back to the **front** of the queue via a new `unshift()` method on the built-in task queues
- `_distributeTask()` no longer re-queues tasks itself; the caller puts them back in the right place

This follows the approach discussed in #1032, with the `stricterFIFO` flag the maintainer asked for before merging.

## Usage

```js
const pool = new Piscina({
  filename: './worker.js',
  minThreads: 1,
  maxThreads: 1,
  stricterFIFO: true,
});
```

## Tests

```
npm run build && npm run test
151 tests, 149 passed, 0 failed
```

New coverage:

- `test/queues.test.ts` — `unshift()` on `FixedQueue` and `ArrayTaskQueue`
- `test/stricter-fifo.test.ts` — 20 concurrent tasks on a single worker finish in submission order when `stricterFIFO: true`

## Benchmark

`npm run benchmark:simple:default` — no regression on the default (non-FIFO) path:

| | ops/sec |
|---|---|
| default | ~7,200 |
| customized | ~8,300 |

(unchanged from baseline on this machine)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -412,10 +412,6 @@ class ThreadPool {
         continue;
       }
 
-      if (this.options.stricterFIFO) {
-        this._returnUndistributedTask(taskInfo);
-      }
-
       if (this.workers.size < this.options.maxThreads) {
         // We spawn if possible
         // TODO: scheduler will intercept this.
@@ -465,16 +461,13 @@ class ThreadPool {
       return true;
     }
 
-    if (!this.options.stricterFIFO) {
-      this._returnUndistributedTask(task, /* toTail */ true);
-    }
-
+    this._returnUndistributedTask(task, !this.options.stricterFIFO);
     return false;
   }
 
   _returnUndistributedTask (task: TaskInfo, toTail = false) : void {
     if (task.abortSignal != null) {
-      if (toTail || !this.options.stricterFIFO) {
+      if (toTail) {
         this.skipQueue.push(task);
       } else {
         this.skipQueue.unshift(task);
@@ -482,7 +475,7 @@ class ThreadPool {
       return;
     }
 
-    if (toTail || !this.options.stricterFIFO) {
+    if (toTail) {
       this.taskQueue.push(task);
     } else if (this.taskQueue.unshift != null) {
       this.taskQueue.unshift(task);
@@ -595,10 +588,6 @@ class ThreadPool {
     const distributed = this._distributeTask(taskInfo, workers);
 
     if (!distributed) {
-      if (this.options.stricterFIFO) {
-        this._returnUndistributedTask(taskInfo, /* toTail */ true);
-      }
-
       // We spawn if possible
       // TODO: scheduler will intercept this.
       if (this.workers.size < this.options.maxThreads) {
--- a/src/index.ts
+++ b/src/index.ts
@@ -70,6 +70,7 @@ interface Options {
   trackUnmanagedFds? : boolean,
   closeTimeout?: number,
   recordTiming?: boolean,
+  stricterFIFO?: boolean,
   loadBalancer?: PiscinaLoadBalancer,
   workerHistogram?: boolean,
 }
@@ -87,6 +88,7 @@ interface FilledOptions extends Options {
   niceIncrement : number,
   closeTimeout : number,
   recordTiming : boolean,
+  stricterFIFO : boolean,
   workerHistogram: boolean,
 }
 
@@ -122,6 +124,7 @@ const kDefaultOptions : FilledOptions = {
   trackUnmanagedFds: true,
   closeTimeout: 30000,
   recordTiming: true,
+  stricterFIFO: false,
   workerHistogram: false
 };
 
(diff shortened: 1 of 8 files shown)


huntiezz:
Updated the commit:

- Added DCO `Signed-off-by`
- Re-signed with SSH

**For maintainers:** CI is waiting on workflow approval for a first-time contributor — could someone hit "Approve and run" when you get a chance?

Related context: this implements the `stricterFIFO` opt-in flag discussed on #1032, targeting #1031.

src/index.ts
```diff
@@ -464,13 +473,30 @@ class ThreadPool {
       return true;
     }
 
+    if (!this.options.stricterFIFO) {
+      this._returnUndistributedTask(task, /* toTail */ true);
+    }
```
metcoder95: What do you think about:

```suggestion
    if (this.options.stricterFIFO != null) {
      this._returnUndistributedTask(task, !this.options.stricterFIFO);
    }
```

src/index.ts
```diff
@@ -413,7 +416,13 @@ class ThreadPool {
       if (distributed) {
         // If task was distributed, we should continue to distribute more tasks
         continue;
-      } else if (this.workers.size < this.options.maxThreads) {
+      }
+
+      if (this.options.stricterFIFO) {
+        this._returnUndistributedTask(taskInfo);
+      }
```
metcoder95: Regardless of the `stricterFIFO` I believe the distribution method is a good idea, can we make it distribute tasks regardless of the flag state?

src/index.ts
huntiezz: Makes sense — adopted this in `239cc96`.

`_distributeTask()` now always calls:

```ts
this._returnUndistributedTask(task, !this.options.stricterFIFO);
```

So `stricterFIFO: false` (default) re-queues to the tail, and `stricterFIFO: true` re-queues to the front. I also simplified `_returnUndistributedTask()` to rely on the `toTail` argument instead of re-checking the flag internally.

src/index.ts
huntiezz: Agreed — updated in `239cc96`.

`_onWorkerAvailable()` now always routes through `_distributeTask()`, and undistributed tasks are always re-queued from there (tail by default, front when `stricterFIFO: true`). Removed the caller-side `if (this.options.stricterFIFO)` re-queue branches in both `_onWorkerAvailable()` and `run()`.

Tests still pass (`151` tests, `149` passed).

metcoder95 (approved):

huntiezz:
Rebased onto latest `current` to clear the merge conflict from #1062.

Locally after the rebase: `npm run build && npm run lint && npm run test` — all green (147 passed, 0 failed).

CI is waiting on workflow approval for the fork push — could someone hit **Approve and run** when you get a chance?

